PR #5, opened 2025-08-28
Add Proof of Concept for E2E Tests

mschluer:
# Summary

As ordered, this contains a first Proof of Concept for End to End testing via User Interface.

# Changes

Added a test in the cypress-framework to depict structure, approach and coding standard.

# Caveats

- The demonstrated pipeline works within a gitlab-setup and is not compatible with github-actions.
- I used my own container registry & repo to showcase pipelining- and branching-strategy. As you are using main instead of master, this would need adjustments.
- I used my own infrastructure to register the user. The user account should be moved into your control.
- As your development environment will most likely utilize tokens, credentials for test accounts should be kept secret. Not to mess with yours I changed the ones in the account – We should aim for a sustainable solution for this challenge.
- The top level code has not been touched. It might be necessary to exclude things from the newly introduced subdirectory.

# Feedback Needed

Please have a close look on Coding-Style, Linting rules and the structure itself as this lies the foundation of future work and hence has to fit your needs.


ayunis-core-e2e-ui-tests/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+.idea/
+.env
```
devbydaniel, line 3: Should we add the `screenshots` directory here as well?

ayunis-core-e2e-ui-tests/.gitlab-ci.yml
devbydaniel, line 1: I guess we can remove this for now

ayunis-core-e2e-ui-tests/cypress.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'cypress';
+
+export default defineConfig({
+	e2e: {
+		baseUrl: process.env['CYPRESS_BASE_URL'] || 'http://localhost:3000',
```
devbydaniel, line 5: I'm getting a linter error here that `process` is not defined. This change in `tsconfig` seems to fix it, does this make sense for you?
image

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
```diff
@@ -0,0 +1,5 @@
+export class E2eTest01 {
```
devbydaniel: Are these test accounts defined per test? What about defining an account once and then reuse with cleanup / delete after each test / test suite?

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
```diff
@@ -0,0 +1,5 @@
...
+	static username: string = `E2E Test 01`;
+	static email: string = `test.accounts@maximilian-schluer.de`;
+	static password: string = `L0caboo-Test`;
```
devbydaniel: Env variables?

ayunis-core-e2e-ui-tests/cypress/pages/chats.page.ts
```diff
@@ -0,0 +1,20 @@
...
+
+	// Components
+	get navigation() {
+		return Navigation;
+	}
+
+	// Elements
+
+	// Helpers
+	validateOn = () => {
```
devbydaniel, line 15: Can we move this to a base `Page` class and extend it for each page to avoid duplicating this? Or will this not be present / be different across pages?

ayunis-core-e2e-ui-tests/cypress/support/e2e.ts
devbydaniel, line 1: What is this file for?

ayunis-core-e2e-ui-tests/tsconfig.json
devbydaniel, line 1: As mentioned above, I had to include `node` to the `types` array and `*.ts` to the `include` array to fix the `process` linting error. Am I missing something in my setup or is this expected?

ayunis-core-e2e-ui-tests/.gitignore
mschluer: Makes sense once the directory is there.
One thing to take care of then: That needs to be the absolute path. In case you want to go with Visual Regression Tests (VRTs) – those are screenshots you don't want to exclude :)

ayunis-core-e2e-ui-tests/cypress.config.ts
mschluer: Jep

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
mschluer: Nope. Test accounts will be shared among tests as long as they don't interfere.
Every Test Account needs a little maintenance (for instance if you reset your system and don't seed it with fixtures or the fixtures themselves).

Hence, I try to utilize as few as possible to keep maintenance effort low :)

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
mschluer: For the password, that might be an approach.

ayunis-core-e2e-ui-tests/cypress/pages/chats.page.ts
mschluer: That definitely can make sense. I can introduce it in the next step :)

ayunis-core-e2e-ui-tests/cypress/support/e2e.ts
mschluer: It's a standard file imported before all tests.
Things defined here apply for the entire test suite.

ayunis-core-e2e-ui-tests/tsconfig.json
mschluer: Classic "worked on my machine" – fixed it! :)

mschluer:
Thanks for the thorough Review. That helps a lot! :)

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
devbydaniel: Ahh so the `01` means first account and not account for the first test?

mschluer: Exactly. I'd assume we'll need 3 in total:
Another one to check account settings and a non-admin to check that they can do exactly what you allow them to.

devbydaniel: Ok nice, let's name these users semantically then.

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
devbydaniel: Ok, can you adjust the configs so we don't have these credentials in plain text in here anymore?

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
mschluer: Test user is now named as General (as this one will be taken for most test cases)

ayunis-core-e2e-ui-tests/cypress/data/accounts/E2eTest01.account.ts
mschluer: TEST_USER_PASSWORD is now taken from env. I also added a sanity check prior to the entire suite for env-variables to be defined.

The idea is to give each and every test user the same password so the maintenance effort is as low as possible and as well it's clear to everybody involved how to log in.

devbydaniel:
PR is obsolete as subsequent PR was merged to main directly.

PR #6, opened 2025-08-31
feat: add Create Calendar Event tool and integrate with chat interface

pavellnk:


ayunis-core-backend/src/domain/tools/application/errors/tools.errors.ts
devbydaniel, line 1: Move to and extend `ToolExecutionError` in `application/tools.errors`. This error class has a flag to decide if the error should be shown to the model or not (which we can do in all these cases I guess)

Maybe we could even drop all the error types and just use the error message in `ToolExecutionError` to specify what went wrong. Keep in mind that these are mostly for the AI model to figure out what it did wrong when using the tools, so the error messages should be something like "you did X wrong, try Y instead".

ayunis-core-backend/src/domain/tools/application/use-cases/generate-ics/generate-ics.use-case.ts
devbydaniel, line 1: I think we should either move this into a separate module in the common folder / as submodule of the tool module or maybe even to the frontend. If this is really only generating a text, we might not even need the whole controller & use case logic.

devbydaniel (changes requested):

ayunis-core-backend/src/domain/tools/application/use-cases/generate-ics/generate-ics.use-case.ts
pavellnk: Yeah, makes sense, I’ll move it to the frontend.

PR #7, opened 2025-09-04
Representative Tests & Pipeline

mschluer:
# Summary

This introduces representative tests, corresponding testIDs and a basic pipeline to run them.
Also, a bug in the Documentation was fixed.

# Changes

- Added Tests, Page Objects and corresponding TestIDs.
- Added a basic pipeline (Secrets missing, see caveats).
- Fixed a bug in the documentation where an endpoint was named wrongly.

# Caveats

- The pipeline is currently not runnable as Secrets and Model Configuration are missing. I do not have the permission to write Secrets, nor do I know exactly which ones are useful. Also, the model configuration is something you might want to cache in some way to save Credits.
- The pipeline could not be tested as the action can run as soon as it's merged into main. This is a security mechanism github uses so that a collaborator like me cannot exhaust your credits.
- The tests were not entirely testable locally, as the testIDs are not on Integration, yet and testing the chat functionality itself did not work lacking API-Keys (and Cap).
- Initially, it was planned to provide one test case. As models tend not to name Chats correctly and as well they cannot be renamed, I decided to split it into two.
- The documentation of the product is partially incomplete. This is conducted in the internal report I will send over.

# Findings

Conducted in the internal report as well.


mschluer:
BTW: Apparently, the pipeline is executed (although there are secrets missing). I strongly recommend restricting this not to exceed your quota.

ayunis-core-e2e-ui-tests/cypress/pages/components/chat-dropdown.component.ts
devbydaniel, line 1: Same as with Pages, we could have an abstract `BaseComponent` with an abstract `rootLocator` property if this is present everywhere

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+		cy.login(E2eTest01.email, E2eTest01.password, E2eTest01.username);
+
+		ChatsPage.chatInput.modelSelectTrigger.should("contain.text", "Claude Sonnet 4")
+
+		ChatsPage.chatInput.textareaChatInput.type(`Name this chat ${timestring} and print the name of the model I am talking to`);
+		ChatsPage.chatInput.submitButton.click();
+
+		cy.reload(); // Has to be removed once the Sidebar updates itself
+		// Make sure not to fail the test just because the external communication is slow
+		cy.withTimeout("XLONG", () => {
```
devbydaniel: What's your reasoning behind using a manual timeout instead of `cy.waitUntil(() => ...)` or `cy.get(..., { timeout: xxx })`? Can we just use cypress built in waiting methods?

ayunis-core-e2e-ui-tests/cypress/pages/base.page.ts
```diff
@@ -0,0 +1,14 @@
+export class BasePage {
```
devbydaniel: should be abstract

ayunis-core-e2e-ui-tests/cypress/pages/base.page.ts
```diff
@@ -0,0 +1,14 @@
...
+	// Properties
+	path = `/`;
```
devbydaniel: Should be abstract without value

ayunis-core-e2e-ui-tests/cypress/pages/chats.page.ts
```diff
@@ -1,20 +1,49 @@
...
+import ModelsDropdown from './components/models-dropdown.component';
+import ChatDropdown from './components/chat-dropdown.component';
+import ConfirmationDialogue from '@pages/components/confirmation-dialogue';
+import Sidebar from './components/sidebar.component';
 
-class ChatsPage {
+class ChatsPage extends BasePage {
 	// Properties
-	path = `/chats`;
+	override path = `/chats`;
```
devbydaniel: override not necessary if base class & this property is abstract

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
devbydaniel, line 1: Some general remarks / questions about this test
- The test should be independent from the output of the model (as you already pointed out, it's neither possible nor relevant to this test to ensure 'correct' model output). So rather than testing for specific message content and a specific thread title we should test that in the chat window there is one message more which is an assistant message and contains any text, and in the sidebar there is one thread more which is not 'Untitled'. Do you think that makes sense?
- Can we split the individual cases so that each case is one action? Your first test case is basically three cases in one which is (1) when I send a message on the `/chat`, I create a new thread, (2) when I send a message to the AI, I get a response and (3) I can delete a thread. I realize that it's probably tricky to split (1) and (2) via a regular user flow, so I guess we need a seeded thread for this?

ayunis-core-e2e-ui-tests/cypress/pages/chats.page.ts
devbydaniel, line 1: I think this page is actually two pages. There is the page on `/chats` which is basically just the chat input and then there is `/chats/:id` which is the actual chat window. Is it intentional that you combined these two?

devbydaniel (changes requested):

ayunis-core-e2e-ui-tests/cypress/pages/components/chat-dropdown.component.ts
mschluer: Done

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
mschluer: In general, you can. The element needs to be accessible though (and this is what we're providing more tolerance for).
This would then mean to duplicate the accessor-method as a getter must not have parameters.
Which would mean duplicating the maintenance effort for this locator.

Another option would be to take the locator directly into the test case which reduces maintainability: You introduce POMs so that the test only describe the semantic of the interactions rather than how to find the elements.
Hence, I decided against this option as well.

As I do not want to break the scheme, I introduced a method to deliberately override the timeout without introducing a magic literal as this one directly depends on the initially provided one.

ayunis-core-e2e-ui-tests/cypress/pages/chats.page.ts
mschluer: Yes, you / we can do that. I'd recommend doing that as soon as we need that.
For the current test, this is not a needed functionality. Every complexity needs to pay for something – here that wouldn't be the case in the current setup.

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
mschluer: **To 1)**

This test verifies not only that it's possible to send messages. As well it verifies, that the message reaches the correct model. As this is crucial to your value proposition, I strongly recommend to check something like this. A fuzzy match should be enough – but also required. If that does not work for some reason and instead of Llama your users communicate with ChatGPT, that would have GDPR Implications. This is the closest possible; maybe with a more robust prompt we could be more reliable there (which according to my checks is not needed).
Do you have an Idea how we could achieve that?

**To 2)**

This is a conscious decision.
When creating E2E-Tests, you try to create as much coverage as _reasonable_ per test case as they are expensive. As you've seen already, even logging in takes 1-2 seconds.

We have a tradeoff here which I would take for coverage. E2E-Tests are your last resort and in the best case should not fail at all: Your previous test stages should have recognized all issues before (in the ideal setup)

I would recommend fine-granular tests like you're proposing for the earlier test stages:
They are intended to help you debug things. E2E-Tests are intended to make sure that you broke nothing your previous stages didn't catch so far.

devbydaniel: To 1)

I 100% agree with your point about testing that the messages reach the correct model. My concern is that this test heavily relies on the model provider and their correct functioning of service which we cannot control and don't want to test. If the model provider is offline or simply takes a very long time to respond (which is common), this test will fail even though our implementation might be correct. 

On first thought, I lean towards creating a mock service that reliably returns what you expect in this test. This is probably a larger initiative, but we need something like this anyway as soon as we want to test the `execute-run` use case, which is the most important use case in the repo. Do you think that makes sense?

I'll try to hack something together this week so that you can leave the test case as is. 

To 2) Ok makes sense.

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
```diff
@@ -0,0 +1,58 @@
+import ChatsPage from '@pages/chats.page';
+import { E2eTestGeneral } from '@data/accounts/E2eTest01.account';
+import { getTimestring } from '@helpers/time.helper';
+
+describe('Chat Lifecycle', () => {
+	it('allows a user to create, rename and delete a chat', () => {
+		const timestring = getTimestring();
+		cy.login(E2eTestGeneral.email, E2eTestGeneral.username);
+
+		ChatsPage.chatInput.modelSelectTrigger.should('contain.text', 'Claude Sonnet 4');
```
devbydaniel: We should probably actively select this model before we send a message to it, right? Or we could store this as the default model for the test user, but I'd actually prefer actively selecting it to make the test more explicit.

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+
+		ChatsPage.chatInput.textareaChatInput.type(
+			`Name this chat ${timestring} and print the name of the model I am talking to`,
+		);
+		ChatsPage.chatInput.submitButton.click();
+
+		cy.reload(); // Has to be removed once the Sidebar updates itself
+		// Make sure not to fail the test just because the external communication is slow
+		cy.withTimeout('XLONG', () => {
```
devbydaniel, line 26: Regarding the [comments below](https://github.com/ayunis-core/ayunis-core/pull/7/files#r2330434061) about wether or not the `ChatPage` is one or two pages. If we ...
- add a new page `NewChatPage` which does not have getters for messages
- use that page above this line 
- add a `ChatPage.validateOn()` here, 

... the pages are correctly reflected and the test case better expresses what actually happens, right?

I would actually prefer that because it's more explicit what happens and the behavior of some components inside the chat input is actually different depending on whether they are used on the `NewChatPage` or the `ChatPage`

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
mschluer: Providing a Mock is a good Idea. Keep in mind that you need to smoke-test the model-communication manually then (although I would go with the proposed way).

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
mschluer: We should select it as a default for the test user. This user is not meant to be used by anyone except the test suite and hence, we can (and should) rely on defaults.
Otherwise that would reveal the default is broken.

ayunis-core-e2e-ui-tests/cypress/e2e/chat-lifecycle.cy.ts
mschluer: Done

devbydaniel:
@mschluer 

as discussed I just added a new commit for seeding data.

The `package.json` of the backend now contains `npm run seed` (for App with Build) and `npm run seed:ts` (for App without Build), to seed data.

I ended up creating a new CLI module where I now use the repositories which turned out to be easier than raw SQL. The values are currently hard coded in [seed-minimal.command.ts](https://github.com/ayunis-core/ayunis-core/pull/7/files#diff-1a25be3690a2de2df1f60a25c9e2ff9f3e6b8623b13d9b2f09aa36df82e5affb). If you want / need them parameterized I can extend this easily, but I thought that for simplicities sake it should work fine for now as is.

You should be able to do something like `docker compose -f docker-compose.yml -f docker-compose.test.yml exec api npm run seed` to generate seed data.

mschluer:
# Updates

- Test Environment is now taken from a compose file in the root directory. This as well as the Dockerfile had to be duplicated:
  - Overriding the compose file while eliminating the previous requirement for a .env-file is not possible.
  - The Dockerfile hard codes the NODE_ENV to production which prohibits seeding
- Minor Adjustments in the tests to respect seeding data.

# Blockers

- Seeding exits improperly with code 137.
- The test account is a basic account.
- The Mock does not reply which lets the test fail (as nothing could be checked)

# Caveats

- Cursor reveals a bug in the application code.
- The Seeding knows too much. Seeding data should be taken from a file [can be done later]
- The setup is quite greedy; caching is heavily recommended [can be done later]

devbydaniel:
PR was merged directly to main

PR #100, opened 2026-01-30
refactor: update LongChatWarning component and add Alert component

chrismahlkeui:
- Replaced Item component with Alert component in LongChatWarning for better UI consistency.
- Introduced new Alert component with AlertTitle and AlertDescription for improved alert handling.
- Updated CSS variables for warning colors in core.css.
- Changed registry URL in components.json to point to local development server.

**NOTE: Button in the alert is not correct for now. l need to change that in the ui library.**

### DIFF
--- a/ayunis-core-frontend/components.json
+++ b/ayunis-core-frontend/components.json
@@ -5,7 +5,7 @@
   "tsx": true,
   "registries": {
     "@ayunis": {
-      "url": "https://registry.ayunis.de/r/{name}.json",
+      "url": "http://localhost:1998/r/{name}.json",
       "headers": {
         "Authorization": "Bearer ${AYUNIS_REGISTRY_TOKEN}"
       }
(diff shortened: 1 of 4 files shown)


ayunis-core-frontend/src/styles/themes/core.css
devbydaniel, line 1: Are these adjustments intentional?

ayunis-core-frontend/components.json
```diff
@@ -5,7 +5,7 @@
   "tsx": true,
   "registries": {
     "@ayunis": {
-      "url": "https://registry.ayunis.de/r/{name}.json",
+      "url": "http://localhost:1998/r/{name}.json",
```
devbydaniel: This only works if you are running the UI library locally, I think we need to keep this at the remote URL though 😅 (you only need to replace the URL back to the previous one, everything else should be fine)

devbydaniel (changes requested):

ayunis-core-frontend/components.json
devbydaniel: @cursoragent please change the url of the ayunis registry back to `https://registry.ayunis.de/r/{name}.json`

devbydaniel (approved):

PR #295, opened 2026-03-02
revert: update UI components for better structure and styling"

chrismahlkeui:
Reverts ayunis-core/ayunis-core#294


chrismahlkeui:
NOTE: The badge has still custom classes. I will change that as soon as Paul accepted my pull request regarding the updated library

PR #618, opened 2026-04-23
fix(models): persist user default model across permitted-model changes (AYC-63)

ayunis-ai:
## Summary

Fixes AYC-63: User default model (Account Settings → Chat) vanishes over time, falling back to the org default.

Root cause: `user_default_models.modelId` referenced `permitted_models.id` with `ON DELETE CASCADE`. Whenever an admin deletes and recreates a permitted model for the same catalog model, every user's preference for that model is silently wiped.

Fix: Reference the catalog model (`models.id`) instead of the permitted-model record. The preference now survives permitted-model re-creation. When the user looks up their default, we resolve the catalog id against the effective (org-wide) permitted set and gracefully fall back to the org default if the catalog model is no longer permitted.

## Changes

- **Migration** — `1776869182118-UserDefaultModelReferenceCatalog.ts`: swap FK from `permitted_models.id` → `models.id`, rewrite existing rows via join, drop orphans.
- **Schema** — `user-default-model.record.ts`: `@ManyToOne(() => LanguageModelRecord, { onDelete: 'CASCADE' })`.
- **Port + use cases** — `GetUserDefaultModelUseCase` now takes `orgId`, resolves through `GetEffectiveLanguageModelsUseCase`, returns `null` when the preferred catalog model is no longer permitted. `SetUserDefaultLanguageModelUseCase`, repository, and mapper updated for the new shape.
- **Dead code** — `DeleteUserDefaultModelsByModelId` use-case removed (no callers; cascade is now handled at the DB level via the FK).
- **Tests** — unit test for `GetUserDefaultModelUseCase` covering the effective-set resolution.

## Test plan

- [x] `npm test` in backend — all tests pass
- [x] `npm run lint`
- [x] Backend TypeScript typecheck
- [x] Pre-commit: ESLint, Prettier, complexity, depcruise, duplication, file-size, markdownlint
- [ ] Manual QA: create user default → admin deletes+recreates the permitted model → user default survives
- [ ] Manual QA: migration runs cleanly on staging against existing data

Refs: AYC-63


ayunis-ai:
* **#618** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #850, opened 2026-06-17
fix(invites): allow re-inviting a deleted user's email (AYC-299)

devbydaniel:
Deleting a user tried to clean up that user's invite but looked it up
via findOneByEmail, which filtered acceptedAt IS NULL and so never
matched an already-accepted invite. The accepted invite row was left
orphaned, and because invites.email is globally unique, re-inviting the
same email failed on the unique constraint.

- Drop the acceptedAt filter from LocalInvitesRepository.findOneByEmail
  (its only callers are deletion paths; email is globally unique so at
  most one row matches)
- Add a data-cleanup migration removing accepted invites with no
  matching user (unblocks already-affected customers)
- Add a regression test for findOneByEmail

---

> [!NOTE]
> **Medium Risk**
> Changes invite lookup behavior used on deletion paths and runs irreversible DELETE migration on production invite data; scope is narrow and logic is well documented.
> 
> **Overview**
> Fixes **AYC-299**, where deleting a user could leave an **accepted** invite row behind and block re-inviting the same email because `invites.email` is globally unique.
> 
> **`LocalInvitesRepository.findOneByEmail`** no longer filters to pending invites only (`acceptedAt IS NULL` removed). User-delete and delete-by-email flows use this lookup to remove the invite row regardless of acceptance state.
> 
> A **one-way data migration** deletes accepted invites whose email has no matching user (case-insensitive), clearing rows created before the code fix without touching accepted invites for users that still exist.
> 
> **Regression tests** assert `findOneByEmail` returns accepted invites and does not require `acceptedAt` in the query.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93d1856b383b5bd2714a6b49f47da08e971c1745. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ayunis-core-backend/src/db/migrations/1781707812150-CleanupOrphanedAcceptedInvites.ts
+++ b/ayunis-core-backend/src/db/migrations/1781707812150-CleanupOrphanedAcceptedInvites.ts
@@ -1,0 +1,36 @@
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * Data cleanup for AYC-299: deleting a user failed to remove that user's
+ * already-accepted invite, leaving an orphaned row behind. Because
+ * `invites.email` carries a global UNIQUE constraint, the orphan blocked
+ * re-inviting the same email (the insert hit the unique violation).
+ *
+ * The code path is fixed separately (delete-by-email now matches accepted
+ * invites too), but rows orphaned before that fix still need clearing. This
+ * deletes every accepted invite whose email no longer maps to a user —
+ * i.e. invites left behind by a deleted user. Accepted invites for users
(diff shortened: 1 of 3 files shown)


devbydaniel:
* **#850** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/850?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/850?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #880, opened 2026-06-24
feat(rag): parse PDF links via web fetch and crawling (AYC-266)

sirolad:
## AYC-266 — Support PDF parsing via web fetch

Fetching a URL that points to a PDF previously failed: `CheerioUrlRetrieverHandler` rejected any non-HTML/text response with `UnsupportedContentTypeError`. As a result the `website_content` tool errored on PDF links and the knowledge-base crawler silently skipped them.

### What changed
- `CheerioUrlRetrieverHandler` now detects PDF responses and routes the downloaded bytes through the **existing** file-parsing pipeline (`RetrieveFileContentUseCase` → Mistral OCR or `pdf-parse`), returning the extracted text as the page content.
- Detection covers:
  - an `application/pdf` content type, and
  - a generic/absent content type (`application/octet-stream` or none) on a `.pdf` URL — some hosts serve PDFs this way.
- `UrlRetrieverModule` imports `FileRetrieverModule` to inject the use case. No new ports/adapters — this follows the project's cross-module "use the exported use case" convention.

### Both flows are covered
Both the `website_content` tool and the knowledge-base crawler fetch through the same handler, so:
- **Regular web fetches** of a PDF link now return its text.
- **Crawls** index PDFs encountered on same-site links. PDFs expose no anchors, so they are not crawled further.

### Notes
- A PDF parse failure surfaces as the file pipeline's domain error on single fetches; during a crawl the page is logged and skipped (existing per-page behaviour).
- The `application/pdf` allow-through is the only change to the content-type gate; genuinely unsupported types (e.g. `image/png`) are still rejected.

### Tests
Added cases to `cheerio.url-retriever.spec.ts`:
- parses an `application/pdf` response through the file pipeline (text, title, metadata, no links)
- treats an octet-stream `.pdf` URL as a PDF
- still rejects genuinely unsupported content types

Validation: `eslint --max-warnings=0` (incl. complexity config), `tsc --noEmit`, full `retrievers` jest suite, and `madge` (no new cycles) all pass.

### Content-type preflight & download size cap (review feedback)

Addressing @devbydaniel's review on `cheerio.url-retriever.ts`:

- The fetch handler now rejects unsupported content types from the response **headers**, before buffering the body — via an `assertContentType` callback supplied by `RetrieveUrlUseCase`. Content-type policy stays in the application layer; the adapter only enforces it. Previously the full body was downloaded and only then rejected.
- Added a **download size cap** (default 25 MB, overridable via `url.maxDownloadBytes`): a `Content-Length` preflight plus a streaming read that aborts once the cap is exceeded, so pointing the crawler at a huge PDF can no longer OOM the backend. Exceeding it raises `UrlRetrieverContentTooLargeError` (413). The cap applies to all fetched bodies (HTML/text included), not just PDFs.

### Manual QA

URLs that resolve to a PDF are parsed through the file-retrieval pipeline (Mistral OCR or `pdf-parse`) instead of being rejected. Two user-facing flows share the same URL retriever, so both can be checked from the running app — there is no direct fetch endpoint.

| Flow | Where | Route |
| ---- | ----- | ----- |
| `website_content` tool (single fetch) | Chat — `/chats/$threadId`: ask the assistant to read/summarize a PDF link | `POST /api/runs/send-message` |
| Knowledge-base URL crawl | KB detail — `/knowledge-bases/$id` → **Add URL** | `POST /api/knowledge-bases/{id}/urls` (`{ "url": "...", "maxDepth": 0 }`) |

**Expected before/after** (run the old revision vs. this branch):

- **KB, direct PDF URL, `maxDepth: 0`** — before: the source ends **FAILED** (the PDF is rejected as an unsupported content type, and a root failure aborts the crawl); after: the source reaches **READY** with the PDF text indexed.
- **KB, an HTML page linking a same-site PDF, `maxDepth: 1`** — before: the PDF child page is silently skipped; after: its text is indexed alongside the HTML.
- **Chat** — before: the assistant reports it cannot read the link; after: it summarizes the extracted PDF text.

Test inputs: a small text PDF such as `https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf`; for the generic-content-type path, any URL whose path ends in `.pdf` but is served as `application/octet-stream`.

---

> [!NOTE]
> **Medium Risk**
> Changes outbound fetch behavior (SSRF-adjacent crawling), memory/timeout limits, and PDF OCR paths used by chat and knowledge-base indexing; architecture enforcement may surface follow-up refactors for existing adapter violations.
> 
> **Overview**
> Refactors URL retrieval so the Cheerio adapter only **fetches raw bytes** (`fetch` + `parseHtml`) and **`RetrieveUrlUseCase`** owns content-type policy, charset decoding, and routing PDFs through **`RetrieveFileContentUseCase`** (including octet-stream / redirect-to-CDN `.pdf` heuristics). **`UrlRetrieverModule`** now imports **`FileRetrieverModule`** for that cross-module call at the application layer.
> 
> Adds **`url` config** (`URL_RETRIEVER_TIMEOUT_MS`, `URL_RETRIEVER_MAX_DOWNLOAD_BYTES`, default 5s / 25 MB) in **`root-configs`**, with streaming size caps, header-only **`assertContentType`** rejection, timeout through body read, http(s)-only redirects, and **`UrlRetrieverContentTooLargeError` (413)**. **`CrawlUrlUseCase`** redacts query tokens in skip logs.
> 
> Enforces **`adapters-no-use-cases`** in dependency-cruiser and documents the pattern in the backend skill; known violations include pre-existing usage repo imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d99d44d04cb907f2da7e9d8573e1cde9b108f202. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sirolad:
* **#880** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
devbydaniel, line 1: The infrastructure layer of Cheerio should not know about a file retrieval use case. (In general, infrastructure should never import use cases)

Instead, the preflight check if the response is HTML should move to the use case and the use case should then either use the cheerio infrastructure to parse the HTML or delegate to the file retrieval use case to parse the PDF

devbydaniel (changes requested):

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
```diff
@@ -54,17 +52,13 @@ export class CheerioUrlRetrieverHandler extends UrlRetrieverHandler {
...
         response.headers.get('content-type')?.toLowerCase() ?? '';
-      const body = await response.text();
-
-      if (contentType.includes('text/plain')) {
-        return new UrlRetrieverResult(body.trim(), finalUrl, '');
-      }
+      const body = Buffer.from(await response.arrayBuffer());
 
-      return this.parseHtml(body, finalUrl);
+      return { contentType, finalUrl, body };
```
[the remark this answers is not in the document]
devbydaniel: @surajudeen.akande Can you address this comment, I think that is valid. The handler should only load the response if the content type is expected. Plus: We need a size cap when downloading the file so pointing the AI at a huge PDF won't crash our servers because of OOM. (We had these incidents in the past)

ayunis-core-backend/README.md
```diff
@@ -176,3 +176,35 @@ npm run test
 # Generate test coverage report
 npm run test:cov
 ```
+
```
devbydaniel: This is stray AI right?

devbydaniel (changes requested):

ayunis-core-backend/README.md
sirolad: yes it is, its supposed to be a part of the PR description not here

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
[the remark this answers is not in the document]
sirolad: Done in 294999c6.

- **Only load the body when the content type is expected**: the handler now takes an `assertContentType(contentType, finalUrl)` callback (same pattern as the existing `onRedirect` gate). It runs against the response **headers**, before `arrayBuffer()` — so `image/png`, bare `application/octet-stream` on a non-`.pdf` URL, etc. are rejected without downloading the body. The content-type *policy* stays in `RetrieveUrlUseCase` (reusing its `isPdf`/`assertSupportedContentType` logic); the adapter only enforces it, so infra still doesn't interpret content types.
- **Size cap to avoid OOM**: a transport-level limit owned by the handler (alongside the existing timeout). A `Content-Length` preflight plus a **streaming read** that aborts the moment accumulated bytes exceed the cap — the payload is never fully buffered. Default **25 MB**, overridable via `url.maxDownloadBytes`; exceeding it raises `UrlRetrieverContentTooLargeError` (413). Note it applies to all fetched bodies (HTML/text too), not just PDFs.

Added handler tests for reject-before-read, the Content-Length preflight, the streaming over-cap abort, and under-cap success.

ayunis-core-backend/README.md
sirolad: Yep — good catch. That was manual-QA scratch that shouldn't have landed in the README. Removed it in 294999c6 and moved the same content into the PR description under **Manual QA**, where it belongs.

devbydaniel:
@sirolad Is that ready to be reviewed again? What about the two open bugbot comments? Please fix or resolve them if you think they are not applicable

sirolad:
@cursor push 21f14c64d6

sirolad:
@cursor push b63fae734f

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
```diff
@@ -11,39 +12,48 @@ import {
...
 const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308]);
 const MAX_REDIRECTS = 5;
+// Upper bound for a caller-supplied `options.timeout`, so a request-controlled
+// value can never disable the abort timer that bounds the whole fetch.
+const MAX_TIMEOUT_MS = 60_000;
+const ALLOWED_PROTOCOLS = new Set(['http:', 'https:']);
+// Hard cap on how many bytes we buffer from a fetched URL. Pointing the crawler
+// at a very large PDF (or any oversized payload) would otherwise buffer the whole
+// response in memory and risk an OOM across shared infrastructure (AYC-266).
+const DEFAULT_MAX_DOWNLOAD_BYTES = 25 * 1024 * 1024; // 25 MB
```
devbydaniel: This is bascially dead code. You already configured a default value of 25MB in the config, so this value will never apply

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
```diff
@@ -11,39 +12,48 @@ import {
...
+  UrlRetrieverContentTooLargeError,
 } from '../application/url-retriever.errors';
 import { ApplicationError } from 'src/common/errors/base.error';
 
 const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308]);
 const MAX_REDIRECTS = 5;
+// Upper bound for a caller-supplied `options.timeout`, so a request-controlled
+// value can never disable the abort timer that bounds the whole fetch.
+const MAX_TIMEOUT_MS = 60_000;
+const ALLOWED_PROTOCOLS = new Set(['http:', 'https:']);
```
devbydaniel, line 25: Any reason why you split the config values between `UrlConfig` and this use case? Would it make sense to merge them into a `WebFetchConfig` instead or simply inline all of them here?

devbydaniel (approved):
Please check my comments and fix if they make sense  to you or merge right away if I missed something

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
sirolad: Good catch — removed. The `url` config factory already applies the 25 MB default (via `parsePositiveIntWithDefault`), so the local `DEFAULT_MAX_DOWNLOAD_BYTES` fallback was dead. The handler now reads the value with `configService.getOrThrow<number>('url.maxDownloadBytes')` (same for `url.timeout`), making the config the single source of truth for those defaults.

ayunis-core-backend/src/domain/retrievers/url-retrievers/infrastructure/cheerio.url-retriever.ts
sirolad: The split is deliberate along one axis: operational vs. security. `timeout` and `maxDownloadBytes` are operational limits an operator may legitimately want to tune per deployment (a slower network, a bigger PDF budget), so they live in the `url` config namespace behind `URL_RETRIEVER_*` env vars. The remaining local constants — `MAX_REDIRECTS`, `MAX_TIMEOUT_MS` (the clamp that stops a request-supplied timeout from disabling the abort timer), `ALLOWED_PROTOCOLS`, `REDIRECT_STATUSES` — are security/protocol invariants that must not be weakenable at runtime, so they stay hardcoded in the handler.

With the dead `DEFAULT_MAX_DOWNLOAD_BYTES` fallback now removed, there's no longer a duplicated default between the two places — the `url` factory is the single source of truth for the tunables. I kept the namespace named `url` (rather than renaming to `WebFetchConfig`) since it's already documented in `config/SUMMARY.md` and the env vars are `URL_RETRIEVER_*`.

PR #889, opened 2026-06-25
fix(chat-input): keep attachments in a scrollable area (AYC-339)

chrismahlkeui:
## Problem

Attaching many documents at once made the attachment list grow without bound and push the chat input off-screen — in Chrome the view jumped to the top and couldn't be scrolled back. Reported by Stadt Nürtingen.

## Fix

The attached-sources list and the pending-images row now render inside a small `ScrollFadeContainer`:

- height caps at ~2 rows, then scrolls — it grows with the content, so 1–2 files don't reserve empty space
- a soft bottom fade that appears only while there's more to scroll and disappears once you're at the bottom

## Why a native scroll container and not the shadcn `ScrollArea`?

`ScrollArea` is built for **externally-fixed heights** (`h-full`, `h-[500px]`, as used elsewhere in the repo) — its Radix viewport wraps content in a `display:table` element with a `height:100%` viewport, so it ignores a **content-driven `max-height`**. This strip needs the opposite: shrink with few chips, cap with many. Forcing `ScrollArea` here would mean either reserving empty height (fixed height) or hacking Radix internals via descendant selectors. A native `overflow-y-auto` container is the idiomatic fit; only the fade (no library equivalent) stays custom.

## Test

New chat → attach many documents → the list stays at 2 rows and scrolls, the chat input stays visible. A fade hints at more below and clears at the bottom.

### DIFF
--- a/ayunis-core-frontend/src/widgets/chat-input/ui/ChatInput.tsx
+++ b/ayunis-core-frontend/src/widgets/chat-input/ui/ChatInput.tsx
@@ -25,6 +25,7 @@ import { useFileDrop } from '../hooks/useFileDrop';
 import { PendingImageThumbnail } from './PendingImageThumbnail';
 import { cn } from '@/shared/lib/shadcn/utils';
 import { SourcesList } from './SourcesList';
+import { ScrollFadeContainer } from './ScrollFadeContainer';
 import { showError } from '@/shared/lib/toast';
 import { MicrophoneButton } from './MicrophoneButton';
 import type {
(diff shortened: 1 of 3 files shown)


chrismahlkeui:
* **#889** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/889?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/889?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

devbydaniel (commented):
@chrismahlkeui Can you please check if the scrollable area for attachments which the AI built looks alright?

devbydaniel (approved):

PR #914, opened 2026-07-01
fix(invites): stop logging full invite JWTs (AYC-387)

devbydaniel:
## Problem

Invite JWTs were being written to logs in full, creating a secrets-in-logs risk. Anyone with log access could extract a valid invite token and accept an invite. This occurred in two places:

- `InvitesController.getInviteByToken` — logged `{ token }` (the raw JWT from the URL path).
- `GetInviteByTokenUseCase.execute` — logged `{ token: query.token }` (the same raw JWT).

## Change

- Both log statements now record `{ hasToken: !!token }` instead of the raw JWT. This keeps the diagnostic signal (a request arrived with/without a token) without leaking the secret, and matches the existing `acceptInvite` logging pattern already used in the same controller and the `AcceptInviteUseCase`.
- Extracted the duplicated invite-delivery logic (build accept URL + send email or return URL) from `create` and `resendExpiredInvite` into a private `deliverInvite` helper. This was required because the pre-commit/CI complexity gate runs on the whole changed file and those two pre-existing methods exceeded the 50-line limit; the extraction brings them back under the limit and removes duplication.

## Verification

- `eslint` passes on both changed files (only the pre-existing, ungated `max-params` warning on the DI constructor remains).
- The gated complexity config (`eslint.complexity.config.mjs`) passes on the controller.
- All pre-commit checks pass (ESLint, Prettier, TypeScript typecheck, complexity, dependency, duplication, file-size).

## Notes

The remaining `logger.error('Invalid invite token', ...)` calls in the invites module only log `error.message` / the JWT library error object, not the token string, so they were left unchanged.

Refs: AYC-387

Linear Issue: [AYC-387](https://linear.app/ayunis/issue/AYC-387/stop-logging-invite-jwts-in-full)

<div><a href="https://cursor.com/agents/bc-96502d9d-aa30-40b3-bc86-3db51f52fb67"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-96502d9d-aa30-40b3-bc86-3db51f52fb67"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/ayunis-core-backend/src/iam/invites/application/use-cases/get-invite-by-token/get-invite-by-token.use-case.ts
+++ b/ayunis-core-backend/src/iam/invites/application/use-cases/get-invite-by-token/get-invite-by-token.use-case.ts
@@ -37,7 +37,7 @@ export class GetInviteByTokenUseCase {
   ) {}
 
   async execute(query: GetInviteByTokenQuery): Promise<InviteWithOrgDetails> {
-    this.logger.log('execute', { token: query.token });
+    this.logger.log('execute', { hasToken: !!query.token });
 
     // Verify and decode the JWT token
     let payload: InviteJwtPayload;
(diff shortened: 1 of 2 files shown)


devbydaniel:
Superseded by #917 (AYC-371), which shipped the identical fix — `deliverInvite` extraction so the invite URL/JWT is never logged, plus the same change in `get-invite-by-token.use-case.ts`. After restacking onto main this branch is empty.

PR #929, opened 2026-07-01
feat(users): cascade user deletion to all dependent data (AYC-382)

devbydaniel:
## Summary

Ensures that deleting a user removes **all** of its dependent data instead of leaving orphaned rows and files behind (data-privacy compliance, AYC-382). Two complementary mechanisms:

### 1. Referential integrity via `ON DELETE CASCADE` foreign keys
Several tables carried a bare `userId`/`user_id` column with **no foreign key**, so user deletion left orphaned rows. Added `@ManyToOne(UserRecord, { onDelete: 'CASCADE' })` relations and a migration adding the FKs for:

- `threads` (→ cascades to messages, artifacts, generated-image records, thread assignments)
- `knowledge_bases`
- `user_system_prompts`
- `mcp_integration_user_configs`
- `user_default_models`
- `artifacts`

The migration also:
- Converts `mcp_integration_user_configs.user_id` from `uuid` → `varchar` (via a `USING` cast, **no data loss**) so its type matches `users.id`.
- Purges any pre-existing orphaned rows before adding each constraint, so it applies cleanly on production databases that already contain orphans.

### 2. Application cleanup for assets the database cannot reach (MinIO + RAG index)
`DeleteUserUseCase` now emits an awaited, pre-delete `UserDeletionRequestedEvent`. Listeners clean up external assets while the owning rows still exist:

- **threads** — `ThreadsUserDeletionRequestedListener` deletes MinIO message-image and generated-image blobs for every thread the user owns. The per-thread storage cleanup was extracted into a shared `ThreadStorageCleanupService` (reused by `DeleteThreadUseCase`).
- **knowledge-bases** — `KnowledgeBasesUserDeletionRequestedListener` deletes the sources of the user's knowledge bases (and their RAG index entries + processing files) via `DeleteSourcesUseCase`. This is required because `sources.knowledgeBaseId` is `ON DELETE SET NULL`, so sources/embeddings would otherwise be orphaned.

Listeners are **best-effort** — failures are logged, never thrown — so a transient storage/index error can never block the user deletion itself; the FK cascade still removes all rows.

## Validation

- `pnpm exec tsc --noEmit` — 0 errors
- `pnpm exec eslint` (+ complexity gate) on changed files — clean
- `pnpm run deps:check` — no violations (the pre-existing grandfathered messages-port import was relocated to the extracted service; net count unchanged)
- `pnpm run test` — **2770 passed / 397 suites**
- Migration verified against a real Postgres 16 + pgvector: runs, reverts, re-runs, and produces **zero drift**. Confirmed via SQL that deleting a user cascades away its threads/knowledge bases/system prompts, and that all six FKs exist with `delete_rule = CASCADE`.

## Notes / follow-ups

- The CLI `users:delete` command deletes via the repository directly (bypassing `DeleteUserUseCase`), so it benefits from the DB cascade but does **not** trigger the MinIO/RAG listeners. Routing it through the use case could be a follow-up.
- `usage` records intentionally retain their rows with `userId` set to `NULL` (existing anonymization behavior) and are out of scope here.

Linear Issue: [AYC-382](https://linear.app/ayunis/issue/AYC-382/user-loschung-auf-alle-abhangigen-daten-ausweiten)

<div><a href="https://cursor.com/agents/bc-2b515b0d-6479-416d-9368-18a4b0b0c8ef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2b515b0d-6479-416d-9368-18a4b0b0c8ef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ayunis-core-backend/src/iam/users/application/use-cases/delete-user/delete-user.use-case.ts
```diff
@@ -51,6 +52,14 @@ export class DeleteUserUseCase {
       );
     }
 
+    // Purge dependent data that database cascades cannot reach (MinIO assets,
+    // RAG index entries) while the owning rows still exist. Awaited so it runs
+    // inside the deletion transaction and before the cascading row delete.
+    await this.eventEmitter.emitAsync(
+      UserDeletionRequestedEvent.EVENT_NAME,
+      new UserDeletionRequestedEvent(userToDelete.id, userToDelete.orgId),
+    );
```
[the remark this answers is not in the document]
devbydaniel: Valid finding, but deliberately not fixed in this PR — the quick fixes all just move the inconsistency around:

- Moving the cleanup after the commit doesn't work as-is: the listeners need the thread/message/source rows to resolve MinIO keys, and after commit the FK cascades have already destroyed them. That's why the event fires pre-delete (documented in the listener docblocks).
- The correct design is collect-keys-inside-txn → commit → best-effort post-commit purge with retry, trading "blobs gone, account alive" for retryable orphaned blobs. That requires an after-commit/outbox mechanism we don't have yet.
- This pattern predates the PR: `DeleteSourcesUseCase` already interleaves vector-index and MinIO mutations inside its own transaction, and org deletion (AYC-383) does no external cleanup at all — deleting an org currently orphans all member blobs and RAG entries without this event ever firing.

Interim risk is small: both listeners swallow their own errors, so the rollback-after-purge scenario requires the row delete itself to fail afterwards. We'll address this properly in a follow-up ticket covering user deletion, org deletion, and source deletion with a post-commit cleanup mechanism.

devbydaniel:
* **#929** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/929?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/929?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ayunis-core-backend/src/domain/threads/application/ports/threads.repository.ts
```diff
@@ -89,6 +89,12 @@ export abstract class ThreadsRepository {
     originSkillId?: UUID;
   }): Promise<void>;
   abstract delete(id: UUID, userId: UUID): Promise<void>;
+  /**
+   * Returns the ids of every thread owned by a user. Used by user-deletion
+   * cleanup to purge per-thread object-storage assets before the thread rows
+   * are removed by the users FK cascade.
+   */
```
pavellnk: @devbydaniel, the whole comment is basically saying "findAllIdsByUserId".

ayunis-core-backend/src/domain/threads/application/services/thread-storage-cleanup.service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Removes all object-storage (MinIO) assets that belong to a thread: image
+ * attachments referenced by its messages and any generated-image blobs.
+ *
+ * Shared by thread deletion and user deletion. Storage errors are logged and
+ * swallowed so a transient failure never blocks the surrounding delete — the
+ * database rows are removed regardless, and an orphaned-object sweep backstops
+ * anything that slips through.
+ */
```
pavellnk: @devbydaniel, this comment also doesn't make sense, and especially here since the service itself is well written and I can understand it with just a glance.

ayunis-core-backend/src/iam/users/application/events/user-deletion-requested.event.ts
```diff
@@ -0,0 +1,21 @@
+import type { UUID } from 'crypto';
+
+/**
+ * Emitted synchronously *before* a user row is deleted, so that listeners can
```
pavellnk, line 4: @devbydaniel, i would say this is a good example of a comment.

pavellnk (approved):

PR #1009, opened 2026-07-10
refactor(models): add PromiseUseCase error handling (AYC-000)

pavellnk:
> [!NOTE]
> **Medium Risk**
> Cross-cutting error-boundary behavior changes how some repository failures are classified (notably create-embedding lookup vs not-found), affecting API error responses until the pattern is rolled out consistently.
> 
> **Overview**
> Introduces **`@HandleUnexpectedErrors`** in `common/decorators/` as the standard error boundary for async use cases: **`ApplicationError`** and transitional **`HttpException`** pass through; everything else is logged under the use-case class name and wrapped in the module’s **`Unexpected*Error`**.
> 
> The **use-case-reference** skill and **common** docs now require this decorator instead of hand-written **`try/catch`** in **`execute()`**. Several **models** use cases are migrated (**get-all**, **get-by-id**, **get-model**, **org default**, **create embedding**), with unit tests for the decorator and **`CreateEmbeddingModelUseCase`**.
> 
> **`CreateEmbeddingModelUseCase`** drops the repository **`.catch()`** that turned lookup failures into **`ModelNotFoundByNameAndProviderError`**—those failures now surface as **`UnexpectedModelError`**. **`GetOrgDefaultModelUseCase`** no longer catches and rethrows raw errors without wrapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fcc2d9908b0a69af69ec3bda81552f14901f0879. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pavellnk:
* **#1009** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1009?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ayunis-core-backend/src/common/filters/application-error.filter.ts
```diff
@@ -49,7 +49,12 @@ export class ApplicationErrorFilter extends BaseExceptionFilter {
 
     const httpException = exception.toHttpException();
     const status = httpException.getStatus();
-    const responseBody = httpException.getResponse();
+    // 5xx messages and metadata can contain raw internals (DB errors, provider
+    // responses) — expose only the error code; details stay in logs/Sentry.
+    const responseBody =
+      status >= 500
+        ? { code: exception.code, message: 'Internal server error' }
+        : httpException.getResponse();
```
pavellnk: This fix is unrelated to the main point of the PR.

devbydaniel:
@pavellnk cc @sirolad

Not fully convinced tbh 😅 – yes, the repetitive error handling is gone, but now we have equally repetitive init ceremony (<- I love that word from AI 😀 ) on the base use case with the additional downside that the init ceremony is less expressive than repetitive error handling.

Is there maybe some kind of decorator magic that we can try so that we can still define `execute` on the concrete use case without any additional stuff?

pavellnk:
@devbydaniel, I don’t fully agree with the AI suggestion either, the setup seems clear enough to me 😊 I didn’t like the decorator idea at first, but I think it fits well here. The main purpose of the base class ware shared error handling, and I’m fine with defining the logger in each use case.

pavellnk:
@sirolad, @devbydaniel, I’ve updated the PR. Please take a look.

devbydaniel:
Nice! If it's a decorator, would you still place it in `common/use-case` or rather in `common/decorators/handle-unexpected-error.decorator.ts` or something?

pavellnk:
@devbydaniel, no strong preference, the decorator is definitely use case specific, but I think it wouldn't hurt to move it to `common/decorators`, and then if we have more use case specific decorators we could group them in a folder.

pavellnk:
@devbydaniel, I think it makes sense to merge this ASAP, the idea that we can eliminate the try/catch won't let me sleep 😄

pavellnk:
@sirolad FYI

devbydaniel (approved):

devbydaniel:
I wonder if we could one shot the adjustment of all use cases with this "swarm" feature of Claude Code

PR #1081, opened 2026-07-21
fix(chat): preserve prompt and images when a run fails (AYC-473)

sirolad:
## Problem

Uploading a JPEG into a chat and asking for a summary can fail mid-run with an error. After the error, the chat loses **both** the uploaded image and the typed prompt, so the user has nothing to retry with. (AYC-473)

## Root cause of the failure (why it happens — the message is fixed in the stacked #1089)

The run fails when the uploaded image's **base64 size exceeds the provider's 5 MB limit**. This is **model-dependent**, which is why it looked intermittent:

- **Bedrock** (e.g. Claude Sonnet 4 🇪🇺) enforces the limit on the base64 payload. A ~4.3 MB phone photo → ~5.8 MB base64 → `400 image exceeds 5 MB maximum`.
- **Anthropic-direct** (e.g. Claude Sonnet 4.5) applies its limit to the raw bytes (4.3 MB < 5 MB), so the **same** photo succeeds.

So larger images work on some models and fail on others. **This PR (the stack base) fixes the data-loss half** — regardless of *why* a run fails, the prompt and image are preserved for retry. The specific, actionable "image too large" message is handled in **#1089** (stacked on top).

## Fix

- Record the in-flight submission (prompt text + image `File` objects) on send — for **both** entry points: continuing an existing chat (`ChatPage.handleSend`) and the new-chat first message (`usePendingMessage` → `onSendStart`).
- Detect failure **centrally** in `useMessageSend`: a `hadError` flag is set by the SSE `error` event **and** by every HTTP/network error (including the 403/429 branches that only toast and never call `onError`), then passed to `onComplete(failed)`, which runs in the `finally` of every non-aborted send.
- On failure, restore via `ChatInput.restoreFailedSubmission`, which only fills **empty** fields — a follow-up drafted while the run was still streaming is never clobbered.
- Clear the stored submission once the send completes.

## Verification

- `tsc --noEmit`, ESLint, prettier, dependency + duplication + file-size checks pass.
- Behavioural QA in a running slot (headless Chrome) with a real 4284×5712 photo of handwritten notes:
  - Summarized successfully on a model within limits.
  - On a run-error (SSE `error` event) the prompt **and** image remain in the input — verified for **both** the new-chat first message and continuing an existing chat.
  - On an injected **403**, prompt + image are preserved (previously lost, since 403/429 never called `onError`).
  - A follow-up typed **during** streaming is **not** overwritten when the run then fails.
  - No horizontal overflow at 375/768/1280px.

### Screenshots / demo

**Happy path** — a JPEG of handwritten notes is summarized:

summary

**On failure the prompt + image are preserved** (new-chat first message; the image thumbnail and text stay in the input for retry):

restored on error

demo

> Media hosted on the `pr-media/ayc-473` branch (not part of this PR's diff); safe to delete after merge.

---

> [!NOTE]
> **Low Risk**
> Scoped UX change in the chat send/restore path with guards against overwriting drafts and cross-thread leakage; no auth, persistence, or API contract changes.
> 
> **Overview**
> **Failed chat runs no longer wipe the user’s prompt and image attachments** — the input is repopulated so they can retry without retyping or re-uploading.
> 
> `useMessageSend` now reports whether a send failed via `onComplete(failed)`, setting a `hadError` flag on SSE `error` events and on HTTP/network failures (including 403/429 paths that only showed toasts before). `ChatPage` keeps the last submission (text + `File`s) for both normal send and the new-chat pending-message path, restores through `ChatInput.restoreFailedSubmission` when `failed` is true, and clears that stash on success, cancel, or thread change so a failure on another thread cannot leak content.
> 
> Restore only runs when the input is still empty (no draft typed during streaming); otherwise users get `errorRestoreFailedSubmission` in EN/DE. Ad-hoc restore logic in `handleSend`’s catch was removed in favor of this centralized completion path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a03357c3b649acdb3af37d6efe7c47f920f13b06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
+++ b/ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
@@ -104,7 +104,6 @@ export default function ChatPage({
 
   const queryClient = useQueryClient();
   const chatInputRef = useRef<ChatInputRef>(null);
-  // In-flight submission, restored into the input if the run fails so the user can retry.
   const lastSubmissionRef = useRef<{ text: string; images?: File[] } | null>(
     null,
   );
@@ -256,9 +255,6 @@ export default function ChatPage({
       console.error('Error in useMessageSend:', error);
       showError(t('chat.errorSendMessage'));
     },
-    // Runs in the finally of every send (success or failure). Restore the
-    // prompt + images on any failure — incl. the 403/429 paths that never
-    // reach onError — then clear the stored submission.
     onComplete: (failed) => {
       if (failed) restoreFailedSubmission();
       lastSubmissionRef.current = null;
--- a/ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
+++ b/ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
@@ -316,8 +316,6 @@ export default function ChatPage({
 
   function handleSendCancelled() {
     abort();
-    // Cancelling is intentional: discard the saved submission so it is neither
-    // restored now (abort skips onComplete) nor left stale for a later failure.
     lastSubmissionRef.current = null;
     setIsStreaming(false);
     setPendingSubmission(null);
--- a/ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
+++ b/ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
@@ -28,7 +28,6 @@ import { useRunErrorHandler } from '../hooks/useRunErrorHandler';
 import { useLetterheadChange } from '../hooks/useLetterheadChange';
 import { usePendingMessage } from '../hooks/usePendingMessage';
 import AppLayout from '@/layouts/app-layout';
-import { AxiosError } from 'axios';
 import type { ChatInputRef } from '@/widgets/chat-input/ui/ChatInput';
 import { useCreateFileSource } from '@/pages/chat/api/useCreateFileSource';
 import { useDeleteFileSource } from '../api/useDeleteFileSource';
@@ -105,6 +104,10 @@ export default function ChatPage({
 
   const queryClient = useQueryClient();
   const chatInputRef = useRef<ChatInputRef>(null);
+  // In-flight submission, restored into the input if the run fails so the user can retry.
+  const lastSubmissionRef = useRef<{ text: string; images?: File[] } | null>(
+    null,
+  );
 
   const [threadTitle, setThreadTitle] = useState<string | undefined>(
     thread.title,
@@ -213,6 +216,12 @@ export default function ChatPage({
 
   const handleError = useRunErrorHandler(thread.id);
 
+  const restoreFailedSubmission = useCallback(() => {
+    const last = lastSubmissionRef.current;
+    if (!last) return;
+    chatInputRef.current?.restoreFailedSubmission(last.text, last.images ?? []);
+  }, []);
+
   const handleSession = useCallback((session: RunSessionResponseDto) => {
     if (config.env === 'development') {
       // eslint-disable-next-line no-console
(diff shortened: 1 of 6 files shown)


sirolad:
* **#1089** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1089?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1081** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1081?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1081?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ayunis-core-frontend/src/widgets/chat-input/ui/ChatInput.tsx
```diff
@@ -111,6 +111,7 @@ interface ChatInputProps {
 export interface ChatInputRef {
   setMessage: (message: string) => void;
   sendMessage: (message: string) => void;
+  restoreFailedSubmission: (text: string, files: File[]) => void;
```
pavellnk: @sirolad, just a question can the restore thing go wrong? maybe it make sense to return a boolean not void?

ayunis-core-frontend/src/widgets/chat-input/ui/ChatInput.tsx
```diff
@@ -222,6 +222,12 @@ const ChatInput = forwardRef<ChatInputRef, ChatInputProps>(
         if (!text.trim() || !modelId) return;
         onSend(text);
       },
+      // Restore only into a pristine input, never onto a mid-run follow-up draft.
+      restoreFailedSubmission: (text: string, files: File[]) => {
+        if (message.trim() || pendingImages.length > 0) return;
```
pavellnk: @sirolad , follow up ot the boolean void. 

```
if (message.trim() || pendingImages.length > 0) {
    return false;
}
.......

return ture
```

ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
```diff
@@ -105,6 +112,9 @@ export default function ChatPage({
 
   const queryClient = useQueryClient();
   const chatInputRef = useRef<ChatInputRef>(null);
+  const lastSubmissionRef = useRef<{ text: string; images?: File[] } | null>(
```
pavellnk, line 114: @sirolad, is it make sense to save it with the thread id? to make sure we wil nto have a cross trread reace?

pavellnk (approved):
A few things definitely need to be checked, but they’re just the things that caught my eye. I could be wrong. :)

ayunis-core-frontend/src/pages/chat/ui/ChatPage.tsx
sirolad: this is already handled on line 142

ayunis-core-frontend/src/widgets/chat-input/ui/ChatInput.tsx
sirolad: It could fail for follow up messages and fail silently, letting it return boolean can give the user a feedback is a better user experience. I have changed that

pavellnk (approved):

devbydaniel:
### Merge activity

* **Jul 23, 8:44 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1081).
* **Jul 23, 8:44 AM UTC**: @devbydaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1081).

PR #1092, opened 2026-07-21
perf(rag): track chunk offsets during splitting instead of searching (AYC-491)

pavellnk:
- carry absolute content offsets through the recursive split so chunk
  positions are exact and computed in O(text length)
- remove the findChunkPosition fallback that rescanned the document per
  chunk character (event-loop freeze on large uploads) and returned
  wrong offsets for repetitive text
- add offset-contract tests: slice invariant, character-split path,
  oversized-segment recursion, and a 1MB perf guard

---

> [!NOTE]
> **Medium Risk**
> Touches core document chunking for embeddings with algorithm and metadata changes, but behavior is heavily covered by new contract and performance tests.
> 
> **Overview**
> **Refactors the recursive text splitter** so chunk **character offsets and line metadata** are computed while splitting, not by scanning the source document afterward.
> 
> A new **`ChunkAssembler`** / **`PositionedChunk`** pipeline carries absolute positions through recursive separator splitting and overlap handling. The old **`findChunkPosition`** heuristics (suffix/middle search per chunk) are removed, which fixes wrong offsets on repetitive text and avoids **quadratic rescans** that could freeze the event loop on large uploads (~1MB perf test).
> 
> **Config handling** is tightened: invalid `chunkSize` / `chunkOverlap` values throw; when only a small `chunkSize` is set, default overlap is capped to a 20% ratio instead of using the global 200 default. Whitespace-only tails (e.g. long tab runs) no longer emit empty chunks.
> 
> Tests add offset invariants, zero-overlap behavior, a **1MB timing guard**, and **300-case** randomized offset-contract checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3f35311d838be3b79d07d86302f89efd5470cec3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1094, opened 2026-07-21
fix(runs): defuse crash vectors in SSE, share events, and cleanup cron (AYC-491)

pavellnk:
- SSE stream: register an 'error' listener (socket errors before
  'close' were uncaught exceptions) and guard the heartbeat write
  which could throw from a timer context and kill the process
- buffered-stream helper: add return()/throw() so a client disconnect
  unsubscribes the provider stream instead of buffering to the end of
  the generation
- delete-share: emitAsync().catch() instead of bare emit(); the async
  listeners do DB writes and a listener failure became an unhandled
  rejection
- stale-processing cleanup cron: wrap the run in try/catch like the
  other cleanup tasks
- decompose sendMessage and move Swagger constants to
  runs.controller.openapi.ts to satisfy the complexity and file-size
  gates on the touched file

---

> [!NOTE]
> **Medium Risk**
> Touches the primary chat SSE path and trial billing side effects; behavior changes around error payloads and disconnect handling need regression on streaming clients.
> 
> **Overview**
> **Refactors `POST /runs/send-message`** into a thin controller, a `SendMessageUseCase` (trial consumption + run execution), validators/mappers, and a dedicated `RunSsePresenter`, with Swagger and image-upload config split out for size/complexity limits.
> 
> **SSE reliability (AYC-491):** `RunSsePresenter` now listens for response `error` events (not only `close`), wraps heartbeat writes in try/catch, and guards error-frame writes so timer/socket paths cannot take down the process. Client disconnect stops forwarding; socket integration tests cover mid-stream abort.
> 
> **Client-facing errors:** `RunEventResponseMapper` drops server-only `stack`/`error` detail keys while keeping `ApplicationError` metadata (e.g. `retryAfterSeconds`). Stream failures use domain codes/metadata instead of leaking stacks in SSE payloads.
> 
> **Trial accounting:** Trial increments move into `SendMessageUseCase` as fire-and-forget with logged `.catch` so failures never block or reject the run stream.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4822cd68dfc905ed4687ae8ad131fbb1b22bf86b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pavellnk:
* **#1171** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1096** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1095** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1119** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1119?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1118** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1118?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1094** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1094?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1094?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1117** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1093** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1093?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1092** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1092?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

pavellnk:
Hey @devbydaniel, @sirolad — quick context on why this PR grew beyond its original scope.

**What the fix is actually about:** two crash vectors in the `send-message` SSE stream, both process-killers under Node's default behavior (part of the AYC-491 stability work):

1. The response socket had no `'error'` listener. A socket error (proxy timeout, client dropping mid-stream — e.g. someone closes their laptop) fires _before_ `'close'`, and an unlistened stream `'error'` is re-thrown by Node as an uncaught exception → the whole process dies → ~40s of 502s for every org on the instance. With a listener attached, Node just delivers the event; we log a warning and close that one stream.
2. The 15s heartbeat wrote to the response from inside `setInterval` — a timer callback runs outside any exception filter, so a write to a destroyed socket was also a fatal uncaught exception. The write is now guarded (`writableEnded` check + try/catch inside the callback).

**Why the diff is big:** the fixes pushed the already-long controller methods over our complexity gates, and instead of just shuffling private methods around I ended up restructuring the whole thing — the controller was doing multipart validation, DTO mapping, trial accounting, SSE transport, event mapping, and error framing in one class (463 lines). It's now 146 lines of decorators + delegation, with the logic in focused, individually tested units:

- `RunSsePresenter` — owns the SSE transport (headers, framing, heartbeat, disconnect/error tracking, teardown). Both crash fixes live here, now with direct unit tests.
- `RunEventResponseMapper` — run event → response DTO mapping.
- `SendMessageRequestValidator` — the cross-field multipart rules DTO decorators can't express.
- `SendMessageUseCase` (new, application layer) — trial-message accounting moved out of the controller; the decision _input_ (subscription context) stays at the HTTP boundary as a `consumeTrialMessage` flag on the command, and the fire-and-forget increment now has a `.catch()` so it can't become an unhandled rejection.

**Three small behavior changes** worth individual attention in review (all verified against the frontend before changing):

- SSE error events no longer leak `error.stack` / `error.toString()` to clients — `ApplicationError` metadata like `retryAfterSeconds` still passes through (the quota toast depends on it).
- Removed the controller's duplicate session event — `ExecuteRunAndSetTitle` already yields the session start, so clients were getting it twice.
- Requests failing validation/input-mapping now get a proper HTTP 400 instead of a 200-with-SSE-error-event (validation runs before headers are sent). Also dropped the injected-but-never-used `HasActiveSubscriptionUseCase`.

Everything else is code moved verbatim. 29 new tests; 141 pass across the runs module.

PR #1173, opened 2026-07-29
fix(anonymize): serve pii analysis off the event loop (AYC-561)

devbydaniel:
- Dispatch /analyze to the threadpool instead of the event loop, which
  synchronous GLiNER inference blocked entirely — the service handled one
  analysis at a time and could not answer /health while busy, so
  concurrent callers queued until the backend's 30s timeout fired
- Bound analysis concurrency to 2; the default threadpool allows 40 and
  working memory grows ~3.3 MB per 1k characters of input
- Give /health its own thread token so a saturated analysis queue cannot
  delay it past the container healthcheck timeout
- Lock the lazy model singleton, which is now entered concurrently; a
  second copy of the model would cost ~1.9 GB
- Reject input beyond 100k characters, which already exceeds the caller's
  30s timeout and holds ~335 MB while it runs
- Pre-cache the mdeberta tokenizer so the first model load no longer
  fetches from huggingface.co inside a request

---

> [!NOTE]
> **Medium Risk**
> Changes request scheduling, concurrency, and input limits for a CPU-heavy inference path; mis-tuning could cause queueing or rejected legitimate large payloads, but behavior is more predictable than blocking the event loop.
> 
> **Overview**
> Fixes **AYC-561** by moving synchronous GLiNER/Presidio work off the FastAPI event loop so `/analyze` no longer blocks other requests (including `/health`) while inference runs.
> 
> **`/analyze`** now runs in a worker thread via `anyio.to_thread.run_sync`, with a dedicated **`MAX_CONCURRENT_ANALYSES`** cap (default **2**, overridable by env). **`/health`** uses a separate one-slot limiter so a full analysis queue cannot stall container healthchecks; it still triggers lazy model load so `service_healthy` only passes when analysis can actually run.
> 
> Adds **double-checked locking** on the lazy `PresidioService` singleton so concurrent first requests cannot load duplicate ~1.9 GB models. Rejects oversized input with **`max_length=30_000`** on request text so work is unlikely to exceed the caller’s ~30s timeout while holding an analysis slot.
> 
> The **Docker image** pre-downloads **`microsoft/mdeberta-v3-base`** tokenizer files at build time so the first in-request model load does not call Hugging Face.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddacc8d8b4fcaec0c7c6cb260fa4adbd0edb564b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ayunis-core-anonymize/Dockerfile
+++ b/ayunis-core-anonymize/Dockerfile
@@ -15,6 +15,13 @@ RUN uv sync --frozen --no-dev
 ENV HF_HOME=/root/.cache/huggingface
 RUN uv run python -c "from huggingface_hub import snapshot_download; snapshot_download('urchade/gliner_multi_pii-v1')"
 
+# The GLiNER checkpoint carries the backbone weights but not its tokenizer,
+# which lives in the mdeberta repo it was trained from — so without this the
+# first model load reaches out to huggingface.co at runtime, inside a request.
+# Only the tokenizer files are wanted (4 MB); the full repo would pull another
+# copy of the backbone weights.
+RUN uv run python -c "from huggingface_hub import snapshot_download; snapshot_download('microsoft/mdeberta-v3-base', allow_patterns=['config.json', 'tokenizer_config.json', 'spm.model'])"
+
 # Copy application code, then re-sync so the project install reflects the
 # actual source (otherwise `uv run` does this on every container start and
 # restarts uvicorn mid-boot, blowing past the healthcheck window).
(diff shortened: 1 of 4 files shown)


devbydaniel:
* **#1173** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1173?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1173?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1194, opened 2026-07-30
ci: switch internal and production deploys to ghcr image pulls (AYC-589)

devbydaniel:
Stage 2 of AYC-589: internal and production stop building on the host, and the release pipeline stops racing the image build. Staging has been running this path since #1193.

Rebuilt on current `main` rather than rebased — see the note at the bottom.

## The part that isn't just "swap build for pull"

`release-please.yml` created the tag, fired `gh workflow run build-images.yml` **fire-and-forget**, and deployed immediately. That was safe only because the deploys built their own images. Once they pull, it's a hard race — on v2.20.1 the deploys started ~18:36 and the images landed 18:42.

So `build-images.yml` gains a `workflow_call` trigger and release-please invokes it as a real job with `needs:`, so a failed image build blocks the deploy instead of shipping into a missing tag. The `actions: write` permission goes away with the dispatch.

A reusable-workflow call runs with the **caller's** ref (`refs/heads/main`), not the tag, which is what `release_tag` corrects:

- **Checkout** takes `ref: ${{ inputs.release_tag || github.sha }}` — `github.sha`, not `github.ref`, since on the push path the branch tip may already have moved past the commit whose `sha-` tag is being published.
- **Tags** — the ref-derived entries are gated off for a `release_tag` build, which would otherwise republish `sha-`/`latest` and defeat the `latest=false` rule. These now live in the `merge` job, since that's where multi-arch manifest lists get tagged.
- **`notify-infra`** was gated on `startsWith(github.ref, 'refs/tags/v')`, false on this path — ayunis-infra would have silently stopped getting the Renovate ping. Now also fires on `release_tag`.
- **Concurrency** was keyed on `github.ref` alone, so a release call and the build from the push that created it shared a group; a third queued run cancels the pending one. Now keyed on `inputs.release_tag || github.ref`.

## Deploy scripts

All four (both release-please jobs, both manual workflows) now match the shape `deploy-staging.yml` has been running in production since #1193 — including everything learned from the staging outage:

- scoped `docker compose pull app code-execution anonymize`, explicit `docker pull` of the sandbox tag, guard aborting before `down`, `up -d --no-build`
- `set -e` retained from #1196 (the ignored `script_stop` input stays gone)
- cleanup narrowed to `ghcr.io/ayunis-core/ayunis-core-*` per #1199, so it can't touch `ayunis-backup`
- `CORE_TAG` from `git rev-parse HEAD | cut -c1-7`, matching metadata-action's unconditional 7-char truncation

Two environment-specific notes:

- **The manual workflows default their `tag` input to `main`**, which is not an image tag, so they resolve `CORE_TAG` from what they actually checked out — the version tag if `git show-ref` finds one, otherwise the `sha-` form.
- **`deploy-internal-manual.yml`'s cleanup moved after `up -d`.** Its unfiltered `docker image prune -af` used to run while everything was down; with pulls that would delete the images just pulled. `docker builder prune -af` stays at the end — the host no longer builds.

All five temporary `docker build -t python-sandbox:latest` lines from AYC-588 are now gone.

## Architecture

Internal and production are both `x86_64`; staging is `aarch64`. #1198 publishes every tag as a `linux/amd64` + `linux/arm64` manifest list, so all three are covered and each host pulls its own platform.

## Verification

- `actionlint` clean across all workflows; all five deploy scripts pass `bash -n`, start with `set -e`, use the narrowed cleanup filter, and contain no `rev-parse --short=7`.
- `git diff main` confirms this branch **adds to** main rather than reverting it: it touches only the four workflows plus `DEPLOYMENT.md`, leaves `ghcr-cleanup.yml` and `deploy-staging.yml` untouched, and the multi-arch `merge` job / `push-by-digest` / arm runners are all still present.
- Real behaviour can only be confirmed on a release. Cheapest rehearsal before that: `workflow_dispatch` `deploy-internal-manual.yml` with `tag: v2.20.1` — but note that tag predates multi-arch and is amd64-only, which is fine for internal (x86_64).

## Why rebuilt, not rebased

The previous revision was auto-restacked and resolved conflicts by reverting `main`: it dropped multi-arch entirely (no `merge` job, no `push-by-digest`, no arm runners), restored the bogus `script_stop`, and removed #1199's and #1200's changes. That revision was discarded and the intended changes re-applied on top of current `main`.

---

> [!NOTE]
> **High Risk**
> Changes production and internal deployment mechanics and release ordering; mistakes could deploy before images exist, break code execution via sandbox image handling, or exhaust disk during pulls.
> 
> **Overview**
> **AYC-589 stage 2:** internal, production, and release-please deploy paths stop host `docker compose build` / local sandbox builds and use `compose.release.yml` with `CORE_TAG` (`vX.Y.Z` on releases, `sha-<7>` for branch/manual deploys).
> 
> **Release pipeline:** `build-images.yml` is callable via `workflow_call` with optional `release_tag`. Release-please invokes it with `needs:` instead of `gh workflow run`, so failed image builds block deploys. Checkout, image tagging, concurrency grouping, and `notify-infra` are adjusted for the reusable path where `github.ref` is `main`, not the release tag.
> 
> **Deploy scripts** (manual internal/production + release-please SSH steps) mirror staging: pre-pull Core services and sandbox before `down`, disk headroom checks, scoped image reclaim (including legacy host-built tags and BuildKit cache), and `up -d --no-build`. Manual workflows resolve `CORE_TAG` from the checked-out tag or commit.
> 
> **Docs:** `DEPLOYMENT.md` states all deployed environments pull from GHCR with multi-arch manifests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a22e3d51e11435200e85b8e3c8fe003902244a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -7,9 +7,13 @@ name: Build & Publish Images
 #   ayunis-core-python-sandbox  — image user code executes in (DOCKER_IMAGE)
 #
 # Pushes to main produce sha-<commit> + latest tags (what staging pulls,
-# AYC-589); release tags produce vX.Y.Z. release-please dispatches this
-# workflow for its tags explicitly (GITHUB_TOKEN-created tags don't fire
-# push-tag workflows) — see release-please.yml.
+# AYC-589); release tags produce vX.Y.Z.
+#
+# release-please calls this workflow instead of pushing a tag that would
+# trigger it: GITHUB_TOKEN-created tags don't fire push-tag workflows, and
+# since AYC-589 its deploys pull these images, so the build has to be a real
+# `needs:` dependency rather than a fire-and-forget dispatch. The caller's ref
+# is then refs/heads/main, not the tag — hence release_tag.
 #
 # Each image is published as a multi-arch manifest list (AYC-589). This is not
 # optional: staging runs aarch64 while internal and production run x86_64, and
@@ -24,13 +28,24 @@ on:
     branches: [main]
     tags: ['v*']
   workflow_dispatch:
+  workflow_call:
+    inputs:
+      release_tag:
+        description: Build this tag instead of the caller's ref, and tag the images with it
+        type: string
+        required: false
+        default: ''
 
 permissions:
   contents: read
   packages: write
 
 concurrency:
-  group: build-images-${{ github.ref }}
+  # release_tag, not just github.ref: a release-please call runs with the
+  # caller's ref (main) and would otherwise share a group with the build
+  # triggered by the very push that created the release — where a third
+  # queued run would cancel it.
+  group: build-images-${{ inputs.release_tag || github.ref }}
   cancel-in-progress: false
 
 jobs:
@@ -69,7 +84,11 @@ jobs:
             target: ''
     runs-on: ${{ matrix.platform == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
     steps:
+      # github.sha, not github.ref: on the push path the ref's tip may already
+      # have moved past the commit whose sha- tag we're about to publish.
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          ref: ${{ inputs.release_tag || github.sha }}
 
       - name: Log in to GHCR
         uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
@@ -179,10 +198,15 @@ jobs:
           # overwrite the main-branch latest that staging pulls.
           flavor: |
             latest=false
+          # A release_tag build runs with the caller's ref (main), so the
+          # ref-derived tags are gated off — otherwise it would republish
+          # sha-/latest and defeat the latest=false rule above. type=semver
+          # still covers a hand-pushed v* tag.
           tags: |
-            type=sha,prefix=sha-
-            type=raw,value=latest,enable={{is_default_branch}}
-            type=semver,pattern=v{{version}}
+            type=sha,prefix=sha-,enable=${{ inputs.release_tag == '' }}
+            type=raw,value=latest,enable=${{ inputs.release_tag == '' && github.ref == 'refs/heads/main' }}
+            type=semver,pattern=v{{version}},enable=${{ inputs.release_tag == '' }}
+            type=raw,value=${{ inputs.release_tag }},enable=${{ inputs.release_tag != '' }}
 
       - name: Create manifest list and push
         working-directory: /tmp/digests
@@ -220,7 +244,12 @@ jobs:
   # and CI_APP_* org secrets shared with this repo.
   notify-infra:
     needs: merge
-    if: startsWith(github.ref, 'refs/tags/v')
+    # On the release-please path github.ref is main, so the ref check alone
+    # would silently stop pinging infra.
+    if: ${{ inputs.release_tag != '' || startsWith(github.ref, 'refs/tags/v') }}
+    # This Renovate wake-up is not part of publishing the images. A transient
+    # notification failure must not fail the reusable workflow and skip deploys.
+    continue-on-error: true
     runs-on: ubuntu-latest
     steps:
       - name: Mint ayunis-infra token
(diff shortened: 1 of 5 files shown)


devbydaniel:
* **#1226** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1226?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1194** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1194?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1194?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1193** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1193?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

devbydaniel:
Valid, and it's a regression this PR introduced — fixed in the latest push.

Confirmed the mechanism: a `uses:` job's conclusion is the aggregate of every job in the called workflow, so `notify-infra` failing would have failed `build-release-images` and skipped both deploys — *after* the images had already published successfully. The Renovate ping is an accelerator (`build-images.yml` L119-124: "so Renovate opens a bump PR ... within a minute or two"), not a release requirement; losing it just means Renovate picks the tags up on its next scheduled run. Blocking a production rollout on it is clearly wrong.

Not theoretical, either: `notify-infra` mints an app token and does `curl -fsS`, both of which fail on any transient auth/API error, and the job depends on the CI app staying installed on `ayunis-infra` plus `CI_APP_*` being shared with this repo. It happened to succeed on the one release run so far ([30571078290](https://github.com/ayunis-core/ayunis-core/actions/runs/30571078290)).

Fix is `continue-on-error: true` on `notify-infra`, which keeps the called workflow's conclusion green while the job still shows as failed in the run. Same fix the autofix branch landed on independently.

devbydaniel:
Valid, and worse than described — fixed in both PRs of the stack, with a different fix than the autofix branch proposed.

Verified both halves:

**The image really is stale.** `--filter until=` reads the image config's `Created`, and a cache-hit rebuild leaves it untouched. The three published `python-sandbox` `sha-` tags — `sha-058178f`, `sha-e7ea22c`, `sha-6155fa7`, built in three separate CI runs across ~40 minutes — all carry `created=2026-07-30T18:04:26Z`, the first build's time. The manifest digests differ (labels), the config timestamp does not. So this isn't an edge case for old tags: once the sandbox Dockerfile is 14 days stale, *every* deploy prunes the image it just pulled.

**There's no self-heal.** `_ensure_sandbox_image()` is called only in the constructor (`executor.py:28`), and both execution paths use `containers.create()` (`executor.py:136`, `executor.py:165`) — which, unlike `containers.run()`, does not pull a missing image. So executions fail with `ImageNotFound` until the container restarts, while `/health` stays green and the deploy reports success. Exactly as described.

**Fix — not the re-pull.** The autofix branch keeps `prune -a` and re-pulls the sandbox afterwards. That re-downloads on every deploy, leaves `-a`'s "delete anything no container references" semantics in place so the same bug returns for any future non-compose image, and the unguarded re-pull sits *after* the container swap. The root cause is reaching for `prune -a` at all, so it's gone: the deploys now delete exactly the `ghcr.io/ayunis-core/*` tags that aren't the one just deployed, which structurally cannot touch what the deploy pulled, plus `prune -f` for dangling. Superseded tags stay in GHCR (`ghcr-cleanup.yml` keeps `sha-*` for 4 weeks, version tags indefinitely), so a rollback re-pulls them.

Also applied to #1193, which had the same bug at `until=72h` — the autofix branch only touched this PR's files.

Verified `docker images --filter "reference=ghcr.io/ayunis-core/*"` matches one path segment (`quay.io/*` matches nothing; `quay.io/*/*` matches `quay.io/minio/minio`), and that the `grep -v`/`xargs -r` pipeline exits 0 under `set -e` when nothing matches.

devbydaniel:
### Merge activity

* **Aug 10, 6:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1194).
* **Aug 10, 6:29 PM UTC**: @devbydaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1194).

PR #1197, opened 2026-07-30
ci: revert staging to build-on-host deploys until images are multi-arch (AYC-589)

devbydaniel:
Reverts #1193. **The staging host is `aarch64`; the published images are `amd64`-only.**

```
ayunis-code-execution-prod  | exec /usr/local/bin/python: exec format error
ayunis-anonymize-prod       | exec /usr/bin/uv: exec format error
```

`build-images.yml` doesn't set `platforms:`, so buildx builds for the GitHub runner's platform — `linux/amd64`. I confirmed from the registry that each tag is a single-platform OCI manifest with `architecture: amd64`, no index. Build-on-host always worked because it built natively on the ARM box.

Reverting restores the build-on-host path (including AYC-588's temporary `docker build -t python-sandbox:latest` line) so the next push to main doesn't break staging again. Re-land once the images are multi-arch.

## Impact of the failed deploy

Staging went down and the workflow still reported success — `anonymize` and `code-execution` crash-looped, so `app` was never created (its `depends_on` requires both `service_healthy`). The 120s health gate returned 124 and that status was discarded, because `script_stop` is not an input of `appleboy/ssh-action@v1.2.5` and has been silently ignored all along. **#1196 fixes that** and would have turned this into a red deploy instead of a green one over a dead stack; it's worth merging regardless of what happens to the GHCR switch.

## Follow-ups this exposes

1. **Multi-arch images.** This repo is public, so free `ubuntu-24.04-arm` hosted runners are available — a per-platform build matrix on native runners plus `docker buildx imagetools create` for the manifest list, rather than QEMU emulation (which would be painfully slow for numpy/pandas and the Nest build). Hosts pull only their own platform, so no extra disk on the boxes.
2. **Check production's and internal's architecture** before #1194 goes anywhere near them, and decide hetzner-03's architecture deliberately in AYC-590 rather than inheriting it.
3. **The image cleanup filter is too broad** — `reference=ghcr.io/ayunis-core/*` matches every image in the org namespace on the host, and the deploy tried to `docker rmi ghcr.io/ayunis-core/ayunis-backup:0.1.23`. It only survived because a running container held the reference. Narrowing to `ghcr.io/ayunis-core/ayunis-core-*` matches just the four Core repos (verified: Docker's `reference` filter supports partial-segment prefix globs). Needs fixing in #1194.

Staging disk is also at 92% (3.2G free), which is what made the *first* attempt fail before any of this surfaced — worth sizing up before retrying the switch.

---

> [!NOTE]
> **Medium Risk**
> Changes how staging is deployed (timing, build vs pull, longer deploys) but preserves abort-on-build-failure before tearing down running containers.
> 
> **Overview**
> **Staging deploys no longer wait on or pull GHCR images** — the workflow triggers on **push to `main`** again instead of after **Build & Publish Images**, and the SSH script builds on the host instead of using `compose.release.yml` and `sha-*` tags.
> 
> The remote deploy path resets to **`origin/main`**, runs **`docker compose build`** (and a local **`python-sandbox:latest`** build) with the same **fail-before-`down`** guard, then **`docker compose up -d`** without `--no-build`. Post-deploy cleanup is simplified to **`docker image prune -f`** instead of removing superseded `ghcr.io/ayunis-core/*` tags.
> 
> **`DEPLOYMENT.md`** now states that published images exist for manual/registry use, but **deploy workflows still build on the host** until registry-based deploys land under **AYC-589** (multi-arch).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b8c2b3779063eca79f724e90a0945ea98e518b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -1,13 +1,8 @@
 name: Deploy to Staging
 
-# Staging runs published GHCR images instead of building on the host (AYC-589).
-# Chained after build-images.yml rather than triggered by the push itself: the
-# image for a commit does not exist until that build finishes (~7 min).
-
 on:
-  workflow_run:
-    workflows: ["Build & Publish Images"]
-    types: [completed]
+  push:
+    branches: [main]
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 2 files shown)


devbydaniel:
* **#1197** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1197?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1197?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

devbydaniel:
Superseded: staging recovered on the multi-arch images (#1198) via the normal pull deploy, so reverting to build-on-host has no purpose.

PR #1198, opened 2026-07-30
ci: publish multi-arch core images for amd64 and arm64 (AYC-589)

devbydaniel:
Root-cause fix for the staging outage. `build-images.yml` never set `platforms:`, so every image was built for the GitHub runner's architecture — `amd64`. Staging runs `aarch64`, so the pulled images died with `exec /usr/local/bin/python: exec format error`. Internal and production are both `x86_64`, so all three environments can't be served by a single-arch image.

Each image is now published as a **manifest list covering `linux/amd64` and `linux/arm64`**.

## Approach

Native runners, not QEMU. The repo is public, so `ubuntu-24.04-arm` is free, and the build legs run on hardware of their own architecture:

```yaml
runs-on: ${{ matrix.platform == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
```

Emulation was the tempting two-line alternative (`setup-qemu-action` + `platforms: linux/amd64,linux/arm64`), but these images compile bcrypt from source, install chromium, and pull torch plus a baked GLiNER model — under QEMU that plausibly turns a 7-minute job into an hour, on every push to main.

The structure is Docker's documented cross-runner pattern:

- **`build`** — 8 legs (4 images × 2 platforms). Each pushes **by digest** with no tags (`push-by-digest=true`), then uploads its digest as an artifact. Tagging per-leg would leave the two architectures overwriting each other's tag.
- **`merge`** — one leg per image. Downloads both digests and binds them into a single tagged manifest list with `docker buildx imagetools create`. This is where `sha-`, `latest` and `vX.Y.Z` are actually applied, so the tag semantics — including the `latest=false` rule — are unchanged and now live in exactly one place.
- **`notify-infra`** now needs `merge` rather than `build`.

The matrix relies on GitHub's documented include-merge rule: an `include` entry whose `image` matches an existing combination contributes its other keys to that combination rather than creating a new job, so each entry supplies `context`/`dockerfile`/`target` to both platform legs. 8 build jobs, not 12.

## Verification

- `actionlint` clean across all 23 workflows.
- The `imagetools create` invocation is the one genuinely fiddly line, so I ran it with stub digests and metadata to confirm the argv:
  ```
  docker buildx imagetools create -t …:sha-28ba734 -t …:latest \
    …@sha256:aaaa1111 …@sha256:bbbb2222
  ```
  The word splitting is deliberate and carries a targeted `shellcheck disable=SC2046` with that rationale.
- A **`Verify both architectures are present`** step fails the job unless `imagetools inspect` reports exactly `linux/amd64,linux/arm64`. Without it, a silently single-arch publish would look identical to success — which is precisely how the outage reached staging.
- All four Dockerfiles were checked for arm64 viability: base images (`python:*-slim`, `node:24.16.0-alpine`, `ghcr.io/astral-sh/uv`) are multi-arch, and the build deps (`apk add python3 make g++`, `chromium`, `apt-get`) plus the numpy/pandas/torch wheels all exist for arm64. `anonymize` is the leg to watch on the first run — it pulls torch and bakes a model.

## Notes

- Hosts pull only their own platform, so this costs the boxes no extra disk. Relevant because staging is at 92% (3.2G free).
- GHA cache scopes go from 4 to 8 (`build-<image>-<platform>`). The repo-wide 10GB cache budget may start evicting; if builds get slower, the first lever is `mode=min` on the arm64 legs.
- Once this lands, the next push to main republishes every tag as multi-arch and staging's existing pull-based deploy should recover on its own — so **#1197 (the revert) can be closed unmerged** if you'd rather fix forward.
- #1194 is still draft and rewrites this file for the `workflow_call` refactor. Its changes now land in different places (tags moved to `merge`, `notify-infra` needs `merge`), so it wants re-applying on top rather than a conflict resolution.
- Unrelated but noticed: `ayunis-core-anonymize/Dockerfile` pulls `ghcr.io/astral-sh/uv:latest` unpinned, which Dependabot can't track.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -10,6 +10,14 @@ name: Build & Publish Images
 # AYC-589); release tags produce vX.Y.Z. release-please dispatches this
 # workflow for its tags explicitly (GITHUB_TOKEN-created tags don't fire
 # push-tag workflows) — see release-please.yml.
+#
+# Each image is published as a multi-arch manifest list (AYC-589). This is not
+# optional: staging runs aarch64 while internal and production run x86_64, and
+# amd64-only images died on staging with `exec format error`. Every leg builds
+# on a runner of its own architecture rather than under QEMU — emulating the
+# torch/chromium/bcrypt builds would turn a 7-minute job into an hour on every
+# push to main. Hosts pull only their own platform, so this costs the boxes
+# nothing.
 
 on:
   push:
@@ -26,11 +34,22 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  # Builds one architecture of one image and pushes it by digest — no tags.
+  # Tagging happens once per image in `merge`, which binds both digests into a
+  # single manifest list. Tagging here instead would leave the two
+  # architectures fighting over the same tag.
   build:
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
+        image:
+          - ayunis-core-app
+          - ayunis-core-code-execution
+          - ayunis-core-anonymize
+          - ayunis-core-python-sandbox
+        platform: [amd64, arm64]
+        # Keys here attach to every combination whose `image` matches, so each
+        # entry covers both platform legs.
         include:
           - image: ayunis-core-app
             context: .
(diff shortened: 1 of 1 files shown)


devbydaniel:
* **#1198** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1259, opened 2026-08-06
fix(tools): sanitize calendar event dates and guard widget render (AYC-675)

devbydaniel:
## Problem

[AYC-675](https://linear.app/ayunis/issue/AYC-675) — the calendar event widget crashed the whole page when the model produced a bad `start`/`end` date. The console showed `RangeError: Invalid time value at Date.toISOString`. The tool params were never sanitized, so a malformed date reached the client and blew up the render. The ask: on wrong parameters, the LLM should receive a helpful error message instead.

## Root cause

- **Frontend crash:** `CreateCalendarEventWidget` parsed `params.start`/`params.end` via `new Date(iso).toISOString()`. For an unparseable value, `new Date(...)` is an *Invalid Date* and `.toISOString()` throws `RangeError`, crashing the widget (and the page) during render — including while streaming partial params.
- **No backend sanitization:** `CreateCalendarEventTool` is a display-only widget tool, so the runtime never executed a handler and its params were never validated. The model got no feedback on bad dates.

## Fix

- **Backend — validate before display.** `CreateCalendarEventTool` now opts into backend execution (`isExecutable`), following the same hybrid displayable+executable pattern as `CreateDocumentTool`/diagram tools. The new `CreateCalendarEventToolHandler` validates the schema and asserts `start`/`end` are parseable ISO 8601 date-times with `end` after `start`. On bad input it throws a `ToolExecutionFailedError` with `exposeToLLM: true` and an actionable message (e.g. `'start' must be a valid ISO 8601 date-time …`, `'end' must be after 'start'`), so the model retries with corrected values. The tool description also spells out the date-time contract.
- **Frontend — never crash on bad dates.** ISO parsing now guards against Invalid Dates (`Number.isNaN(date.getTime())`), falling back to an empty date / default time instead of throwing. This is a defense-in-depth fix so the widget stays robust even for partial/streaming or historical bad params.

## Tradeoff noted

Making the tool executable means it no longer ends the turn immediately (like documents/diagrams, the model gets a display-ack and can add a short confirmation). This is a minor, deliberate behavior change and the idiomatic mechanism in this codebase for feeding a helpful validation error back to the model.

## Tests

- New `create-calendar-event-tool.handler.spec.ts`: valid input resolves; invalid `start`, `end <= start`, and missing required params all reject with an `exposeToLLM` `ToolExecutionFailedError`.
- Full tools + runtime adapter suites pass (201 tests); backend/frontend type-check and lint clean; pre-commit hooks green.

## Acceptance criteria

- [x] Bad calendar params no longer crash the page.
- [x] The LLM receives a helpful, actionable error message on wrong parameters.

Linear Issue: [AYC-675](https://linear.app/ayunis/issue/AYC-675/calendar-widget-bad-input-crashes-the-page)

<div><a href="https://cursor.com/agents/bc-21f40262-6c8a-445c-8736-316ca6812587?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-21f40262-6c8a-445c-8736-316ca6812587&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


devbydaniel:
Closing in favor of #1276 / #1277. The root-cause analysis here was correct and the frontend guard carries over nearly verbatim — thanks. We went a different way on the backend, though: instead of making the calendar tool a hybrid (which costs the instant turn-end and only covers this one tool), display-only tool params are now validated generically in the run loop (`ToolResultCollectorService` in #1276, and a `Tool.validateInput` seam in the agent runtime in #1277). An invalid call feeds the model the validation error and keeps the loop running; a valid call still ends the turn immediately. The calendar date checks (ISO 8601 parseable, end after start) moved into `CreateCalendarEventTool.validateParams`, so every display tool's schema is enforced the same way.

PR #1340, opened 2026-08-11
wip(workspaces): add the workspaces module, feature flag and table (AYC-700)

pavellnk:
Iteration 1 of the Workspaces/Projects plan ([AYC-700](https://linear.app/ayunis/issue/AYC-700)), PR 1 of 5. Backend only — nothing is reachable yet.

Adds a `workspaces` domain module: personal folders that will group a user's chats. A workspace carries a name, an optional description and an appearance (icon key + colour), can be pinned to the sidebar, and holds a manual position in that list.

**User-facing copy says "Projekte"; the code, tables and routes say `workspace`.** The German wording is still open (CS/GTM are being looped in per the ticket comments), so it lives purely in locale files and can change without a migration.

## What's here

- `src/domain/workspaces/` — entity, errors, repository port, 7 use cases, local TypeORM adapter, controller + DTOs. Structure follows `letterheads`; use cases follow the current `@HandleUnexpectedErrors` style rather than letterheads' older hand-written try/catch.
- `CreateWorkspacesTable` migration (generated, not hand-written; drift check is clean).
- `FEATURE_WORKSPACES_ENABLED`, **off by default**, applied at the controller via `@RequireFeature`. Lets iterations 2–6 land on main progressively.

| Method | Route |
| --- | --- |
| POST | `/workspaces` |
| GET | `/workspaces` |
| GET | `/workspaces/:id` |
| PATCH | `/workspaces/reorder` |
| PATCH | `/workspaces/:id` |
| PATCH | `/workspaces/:id/toggle-pinned` |
| DELETE | `/workspaces/:id` |

## Two decisions worth a look

**Pinning and reordering bypass `save()`.** Both go through single-statement SQL in the repository instead of a read-modify-write on the entity. That is deliberate: `updatedAt` is an `@UpdateDateColumn`, and the list page sorts by it — so pinning a workspace or dragging it in the sidebar would otherwise shove it to the top of "Zuletzt aktualisiert". `updateSortOrders` uses `unnest(...) WITH ORDINALITY` so the new order is never partially visible.

**Deletion is announced before it happens.** `DeleteWorkspaceUseCase` emits `WorkspaceDeletionRequestedEvent` *before* the row delete and drains deferred cleanup only after it succeeds — the same two-phase contract as `UserDeletionRequestedEvent`. Nothing listens yet; PR 2 adds the threads listener that purges MinIO blobs, which the FK cascade cannot reach. This also keeps the module dependency one-directional (threads → workspaces), so `madge` and dependency-cruiser stay green.

`PATCH /workspaces/reorder` is declared before `PATCH /workspaces/:id` — reversing them makes `reorder` a 400 from `ParseUUIDPipe`.

## Verification

`tsc --noEmit`, `eslint`, `deps:check`, and 27 new unit tests pass. Migration applied against a local database; re-running `migration:generate` afterwards reports `No changes in database schema were found`.

Refs: AYC-700

---

> [!NOTE]
> **Medium Risk**
> Adds a new persisted domain, migration, and cross-module delete/event wiring; endpoints are off by default but deletion semantics will matter once threads attach to workspaces.
> 
> **Overview**
> Introduces **workspaces** as per-user, org-scoped folders (name, description, icon/color) ahead of chat grouping in later iterations. The capability is gated by **`FEATURE_WORKSPACES_ENABLED`** / `workspacesEnabled` (default **off**), exposed on `GET /feature-toggles`, and enforced on `/workspaces` via `@RequireFeature`.
> 
> A new **`workspaces` domain module** adds a `workspaces` table migration, TypeORM persistence, five use cases (create, list, get, update, delete), validation for appearance keys, and HTTP CRUD. **Create** persists the row then calls **`AddFavoriteUseCase`** so new workspaces show up in the favorites/sidebar model (pin/order stay in **favorites**, not on the workspace row). **Delete** emits **`WorkspaceDeletionRequestedEvent`** before removing the row and runs listener **deferred cleanup** only after delete succeeds; **favorites** registers a listener to drop workspace favorite references (thread/MinIO cleanup is left for a follow-up).
> 
> The frontend picks up **`workspacesEnabled`** (`useIsWorkspacesEnabled`) and regenerated OpenAPI clients for the new workspace DTOs and routes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e30de96b31fe1c5cc05ad096c5c503bbfa3759d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ayunis-core-backend/src/domain/workspaces/application/use-cases/delete-workspace/delete-workspace.command.ts
+++ b/ayunis-core-backend/src/domain/workspaces/application/use-cases/delete-workspace/delete-workspace.command.ts
@@ -1,0 +1,5 @@
+import type { UUID } from 'crypto';
+
+export class DeleteWorkspaceCommand {
+  constructor(public readonly id: UUID) {}
+}
--- a/ayunis-core-backend/src/domain/workspaces/domain/workspace.entity.ts
+++ b/ayunis-core-backend/src/domain/workspaces/domain/workspace.entity.ts
@@ -1,0 +1,65 @@
+import type { UUID } from 'crypto';
+import { randomUUID } from 'crypto';
+import {
+  DEFAULT_WORKSPACE_COLOR,
+  DEFAULT_WORKSPACE_ICON,
+} from './workspaces.constants';
+
+export class Workspace {
+  public readonly id: UUID;
+  public readonly userId: UUID;
+  public readonly orgId: UUID;
+  public readonly createdAt: Date;
+  public name: string;
+  public description: string | null;
+  public icon: string;
+  public color: string;
+  public updatedAt: Date;
+
+  constructor(params: {
+    id?: UUID;
+    userId: UUID;
+    orgId: UUID;
+    name: string;
+    description?: string | null;
+    icon?: string;
+    color?: string;
+    createdAt?: Date;
+    updatedAt?: Date;
+  }) {
+    this.id = params.id ?? randomUUID();
+    this.userId = params.userId;
+    this.orgId = params.orgId;
+    this.name = params.name;
+    this.description = params.description ?? null;
+    this.icon = params.icon ?? DEFAULT_WORKSPACE_ICON;
+    this.color = params.color ?? DEFAULT_WORKSPACE_COLOR;
+    this.createdAt = params.createdAt ?? new Date();
+    this.updatedAt = params.updatedAt ?? new Date();
+  }
+
+  rename(name: string): void {
+    this.name = name;
+    this.touch();
+  }
+
+  describe(description: string | null): void {
+    this.description = description;
+    this.touch();
+  }
+
+  restyle(params: { icon?: string; color?: string }): void {
+    this.icon = params.icon ?? this.icon;
+    this.color = params.color ?? this.color;
+    this.touch();
+  }
+
+  /**
+   * The mapper writes `updatedAt` through to the record, so TypeORM's
+   * @UpdateDateColumn never gets to fill it in — an edit would otherwise keep
+   * its old timestamp and never move in the "last updated" sort.
+   */
+  private touch(): void {
+    this.updatedAt = new Date();
+  }
+}
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -51,6 +51,7 @@ ayunis-core/
 | [anonymization-settings](ayunis-core-backend/src/domain/anonymization-settings) | Privacy Config | Org-level PII whitelist for anonymous mode |
 | [thread-pii-masks](ayunis-core-backend/src/domain/thread-pii-masks/SUMMARY.md) | Privacy | Per-thread PII mask dictionary for anonymous mode |
 | [favorites](ayunis-core-backend/src/domain/favorites/SUMMARY.md) | Favorites | User-owned ordered references resolved for navigation |
+| [workspaces](ayunis-core-backend/src/domain/workspaces/SUMMARY.md) | Folders | Personal folders ("Projekte") that group a user's chats |
 
 ### IAM Modules — Identity & Access Management
 
--- a/ayunis-core-backend/.env.example
+++ b/ayunis-core-backend/.env.example
@@ -262,6 +262,7 @@ AYUNIS_METRICS_PASSWORD=changeme
 # FEATURE_PROMPTS_ENABLED=true
 # Disabled by default
 # FEATURE_SKILLS_ENABLED=false
+# FEATURE_WORKSPACES_ENABLED=false
 
 # Internal Services
 # URLs of the companion services (defaults target the local docker compose stack)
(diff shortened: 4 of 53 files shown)


pavellnk:
* **#1352** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1352?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1351** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1351?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1347** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1347?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1345** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1345?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1344** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1344?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1343** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1343?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1374** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1375** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1375?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1342** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1342?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1340** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1340?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1340?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1367** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1367?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ayunis-core-backend/src/domain/workspaces/domain/workspace.entity.ts
```diff
@@ -0,0 +1,82 @@
...
+export class Workspace {
+  public readonly id: UUID;
+  public readonly userId: UUID;
+  public readonly orgId: UUID;
+  public readonly createdAt: Date;
+  public name: string;
+  public description: string | null;
+  public icon: string;
+  public color: string;
+  public isPinned: boolean;
```
devbydaniel: I wonder if we should model this differently. We might have various different types of entities pinned to the sidebar at some point, so I think I would not put this directly on the workspace.   
  
Also, if a workspace is going to be shared at some point, then multiple people might decide differently on pinning a workspace   
  
So maybe something like `SidebarPin` with `type == pinnedType.WORKSPACE`?

ayunis-core-backend/src/domain/workspaces/domain/workspace.errors.ts
devbydaniel, line 1: this is in the application folder in other modules, I think

ayunis-core-backend/src/domain/workspaces/application/use-cases/delete-workspace/delete-workspace.command.ts
```diff
@@ -0,0 +1,9 @@
+import type { UUID } from 'crypto';
+
+export class DeleteWorkspaceCommand {
+  readonly workspaceId: UUID;
+
+  constructor(params: { workspaceId: UUID }) {
+    this.workspaceId = params.workspaceId;
```
devbydaniel: small nitpick, I would simply put `id` here – it's clear that it's the workspace id

ayunis-core-backend/src/domain/workspaces/domain/workspace.entity.ts
```diff
@@ -0,0 +1,106 @@
...
+    validateWorkspaceDescription(description);
+    this.description = description;
+    this.touch();
+  }
+
+  restyle(params: { icon?: string; color?: string }): void {
+    this.icon = params.icon ?? this.icon;
+    this.color = params.color ?? this.color;
+    this.touch();
+  }
```
devbydaniel, line 55: this is a super nice pattern. We should have done that from the very beginning

devbydaniel (changes requested):

ayunis-core-backend/src/domain/workspaces/domain/workspace.entity.ts
pavellnk: @devbydaniel, sometimes it’s super helpful, but in this case I don’t like it. If the number of fields grows, we’ll need to add more `if` statements, so basically this approach isn’t scalable. But I’ll keep it here for now.

devbydaniel (approved):

sirolad (approved):

PR #1342, opened 2026-08-11
wip(threads): file chats under a workspace (AYC-700)

pavellnk:
Iteration 1 of the Workspaces/Projects plan ([AYC-700](https://linear.app/ayunis/issue/AYC-700)), PR 2 of 5. Stacked on #1340. Backend only — still nothing reachable in the UI.

Connects chats to workspaces: a thread can now be filed under a workspace and pinned in the sidebar.

## What's here

- `threads.workspaceId` (nullable, indexed, `ON DELETE CASCADE`) and `threads.isPinned` on the entity, record, mapper, and both response DTOs.
- `PATCH /threads/:id/workspace` — move a chat into a workspace, or send `null` to take it out.
- `PATCH /threads/:id/toggle-pinned` — returns the new pin state.
- `GET /threads?workspaceId=…` filters to one workspace; `findAll` now orders pinned chats first.
- `POST /threads` accepts an optional `workspaceId`, so a chat can be born inside a workspace.

## Three things worth a look

**Ownership is checked in the use case, not left to the FK.** `AssignThreadToWorkspaceUseCase` and `CreateThreadUseCase` both resolve the target workspace through the workspaces module's exported `FindWorkspaceUseCase`, which is user-scoped. The foreign key alone would happily let you file your chat into someone else's workspace — it only checks that the row exists.

**Deleting a workspace deletes its chats, and their blobs.** The rows go by the FK cascade, but MinIO objects live outside it. `ThreadsWorkspaceDeletionRequestedListener` resolves the affected thread ids *before* the delete (while the rows still exist) and defers the prefix purge until after it succeeds — the same two-phase `DeferredCleanupEvent` contract as the existing user-deletion listener. Lookup failures are logged, never thrown, so they cannot block a deletion. Covered by three tests, including "does not purge when the row delete fails".

**Pinning does not count as editing.** `togglePinned` is a single `UPDATE … SET "isPinned" = NOT "isPinned" … RETURNING` rather than a read-modify-write, so it leaves `updatedAt` *and* `lastActivityAt` untouched. The latter matters beyond cosmetics: data-retention deletion keys off `lastActivityAt`, and a pin bumping it would silently extend how long a conversation is kept.

## Incidental changes

- `GetThreadResponseDto` and `GetThreadsResponseDtoItem` now extend a new `ThreadMetadataResponseDto`. Adding the same two fields to both pushed their already-similar tails over the jscpd duplication threshold; sharing the common fields by inheritance fixes the cause rather than reshuffling lines to dodge the detector, and keeps the two from drifting.
- `LocalThreadsRepository.findAll` had its filter block extracted into `applyFindAllFilters` — the extra condition took it to cyclomatic complexity 11, one over the gate.
- `CreateThreadUseCase`'s spec gained the new `FindWorkspaceUseCase` provider.

## Verification

`tsc --noEmit`, `eslint`, the complexity gate on changed files, `deps:check`, and 163 tests across threads + workspaces pass. Migration applied locally; the follow-up `migration:generate` reports `No changes in database schema were found`.

Refs: AYC-700

---

> [!NOTE]
> **Medium Risk**
> Touches deletion cascades, object-storage purges, and cross-module event ordering; mistakes could leak blobs or stale favorites, but patterns mirror existing deferred-cleanup listeners and are covered by new tests.
> 
> **Overview**
> Adds optional **workspace filing** for conversation threads: nullable `threads.workspaceId` (indexed FK to `workspaces` with `ON DELETE CASCADE`), surfaced on create/list/get responses and via **`PATCH /threads/:id/workspace`** (`null` detaches). **`GET /threads`** accepts `workspaceId`; create accepts optional `workspaceId`. Workspace targets are validated with user-scoped **`FindWorkspaceUseCase`** so chats cannot be filed into another user's workspace despite the FK only checking row existence.
> 
> **Deletion** is extended with a two-phase **`DeferredCleanupEvent`** flow: **`DeleteThreadUseCase`** emits **`ThreadDeletionRequestedEvent`** before the row delete, registers storage purge as deferred work, then runs listener cleanup after delete succeeds. **Favorites** adds **`FavoriteThreadDeletionRequestedListener`** to drop thread favorite rows on direct thread delete. **Workspace delete** adds **`ThreadsWorkspaceDeletionRequestedListener`** to snapshot thread ids, then after cascade purge MinIO prefixes and remove thread favorites for chats that were actually deleted (re-checking survivors so threads moved out mid-flight keep blobs and pins).
> 
> Also adds **`FindThreadsByIdsUseCase`** for batch thread resolution, gates workspace-related HTTP params behind the **Workspaces** feature flag (404 when disabled), and refreshes generated frontend API types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebde0822ce3193eeda44b00d273a5465002586ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pavellnk:
* **#1352** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1352?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1351** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1351?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1347** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1347?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1345** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1345?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1344** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1344?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1343** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1343?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1374** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1375** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1375?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1342** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1342?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1340** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1340?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1367** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1367?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ayunis-core-backend/src/domain/threads/presenters/http/threads.controller.ts
```diff
@@ -39,6 +39,12 @@ import { UpdateThreadTitleCommand } from '../../application/use-cases/update-thr
 
 import { CreateThreadDto } from './dto/create-thread.dto';
 import { UpdateThreadTitleDto } from './dto/update-thread-title.dto';
+import { AssignThreadWorkspaceDto } from './dto/assign-thread-workspace.dto';
+import { AssignThreadToWorkspaceUseCase } from '../../application/use-cases/assign-thread-to-workspace/assign-thread-to-workspace.use-case';
```
sirolad: absolute imports please

pavellnk: Done — every new backend file across the stack (workspaces module in #1340 and the threads additions here) now uses absolute `src/…` imports; the import lines added to existing files are converted too. Same-directory siblings (`./x.command` next to its use case) stay relative, matching how those pairs are imported elsewhere — happy to convert those as well if you'd prefer.

ayunis-core-backend/src/domain/threads/domain/thread.entity.ts
```diff
@@ -16,6 +16,8 @@ export class Thread {
   title?: string;
   messages: Message[];
   isAnonymous: boolean;
+  workspaceId: UUID | null;
+  isPinned: boolean;
```
devbydaniel: Same thing as in the other PR, I'm not sure if we should put this directly on the thread. Maybe at some point, multiple people will be able to make the decision if they want to pin the same thread.   
  
In that case, the `ThreadWorkspacePin` could even live in the workspace module. That way, you also avoid that slightly awkward edge case that `isPinned` might be `true`, but `workspaceId` is `null` which is actually not possible. You simply have `workspaceId` on the thread, the rest is in the workspace module

devbydaniel (changes requested):

devbydaniel (approved):
Just to be sure – the name of the PR is stale right? This PR does not contain any pinning functionality anymore

pavellnk:
> Just to be sure – the name of the PR is stale right? This PR does not contain any pinning functionality anymore

yes, ​stale title

PR #1382, opened 2026-08-12
refactor(iam): join the ambient transaction in provisioning repositories (AYC-627)

sirolad:
## Summary

Routes the three repositories on the SSO provisioning write path through the ambient CLS transaction, so `@Transactional()` actually contains their writes. Scoped precursor to AYC-641 (Phase 4), which requires creating a user, federated identity, invite consumption, and admission result **atomically**.

Converted to the reference pattern from `local-skill.repository.ts:38` (`this.txHost.tx ?? repo.manager`):

- `local-users.repository.ts` — 15 call sites, including two query builders
- `local-invites.repository.ts` — 10 call sites, including two query builders
- `local-subscriptions.repository.ts` — 9 call sites, plus the two `dataSource.transaction(...)` blocks

`users.module.ts` builds its repository via `useFactory`, so `TransactionHost` was added to the factory and its `inject` list.

### The subscriptions case is the interesting one

`create` and `replace` opened their **own** transactions via `this.dataSource.transaction(...)`. Nested inside an outer transaction, those commit independently — an outer rollback would leave the rows behind, which is precisely the failure AYC-627 documents. They now go through:

```ts
private runInTransaction<T>(work: (manager: EntityManager) => Promise<T>): Promise<T> {
  const ambient = this.txHost.tx as EntityManager | undefined;
  return ambient ? work(ambient) : this.dataSource.transaction(work);
}
```

Standalone callers keep their own transaction; nested callers join the caller's. No behaviour change outside a transaction.

## Does NOT close AYC-627

This converts 3 of the ~43 local repositories. Registration still writes through `orgs`, `role-permissions`, `trials`, and `legal-acceptances`, none of which join yet — so registration remains non-atomic and AYC-627 stays open.

## Validation

- `pnpm exec tsc --noEmit` — clean
- `pnpm exec jest src/iam/invites src/iam/subscriptions src/iam/users` — 47 suites, 275 tests passing
- Full suite — 4338 tests passing
- Added tests asserting the observable behaviour: the repository uses the transactional entity manager when one is active, and falls back to the pooled manager when not

### Pre-existing failure, unrelated to this change

Three `rag/embeddings` suites fail to **compile** on this base: Jest cannot parse `p-queue@9.3.1` (`import { EventEmitter } from 'eventemitter3'` — ESM-only). No file in this diff is in that import chain. Likely a recent dependency bump on `main`; worth its own ticket.

---

> [!NOTE]
> **Medium Risk**
> Changes transaction boundaries on the SSO provisioning write path and subscription create/replace; incorrect ambient-tx detection could leave partial commits or break atomicity expectations.
> 
> **Overview**
> **Routes IAM provisioning persistence through the ambient `@Transactional()` CLS transaction** so user, invite, and subscription writes roll back with an outer failure instead of committing on a separate connection.
> 
> `LocalUsersRepository`, `LocalInvitesRepository`, and `LocalSubscriptionsRepository` drop injected TypeORM repositories and resolve `InviteRecord` / `UserRecord` / subscription entities via `TransactionHost` (`txHost.tx.getRepository(...)`). Outside a transaction the adapter still supplies a fallback manager, so standalone calls behave as before.
> 
> **Subscriptions `create` / `replace` no longer always open nested `dataSource.transaction` blocks.** They use `runInTransaction`, which runs on the caller’s manager when `isTransactionActive()` is true and only opens a new transaction when none is active—fixing independent commits that could survive an outer rollback.
> 
> `users.module.ts` injects `TransactionHost` into the `UsersRepository` factory. **Tests** cover transactional vs fallback reads for invites and join-vs-own-transaction behavior for subscription `create`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7746d4592e2d4397d0490736835abcfae56ccf25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ayunis-core-backend/src/iam/invites/infrastructure/persistence/local/local-invites.repository.spec.ts
+++ b/ayunis-core-backend/src/iam/invites/infrastructure/persistence/local/local-invites.repository.spec.ts
@@ -21,13 +21,23 @@ function makeAcceptedInviteRecord(email: string): InviteRecord {
 
 describe('LocalInvitesRepository', () => {
   let inviteRepo: jest.Mocked<Pick<Repository<InviteRecord>, 'findOne'>>;
+  let txHost: {
+    tx: { getRepository: jest.Mock };
+    isTransactionActive: () => boolean;
+  };
   let repository: LocalInvitesRepository;
 
   beforeEach(() => {
     inviteRepo = { findOne: jest.fn() };
+    // Reads resolve through txHost.tx, which outside a transaction is the
+    // adapter's fallback manager rather than undefined.
+    txHost = {
+      tx: { getRepository: jest.fn().mockReturnValue(inviteRepo) },
+      isTransactionActive: () => false,
+    };
     repository = new LocalInvitesRepository(
-      inviteRepo as unknown as Repository<InviteRecord>,
       new InviteMapper(),
+      txHost as never,
     );
   });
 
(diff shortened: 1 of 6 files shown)


sirolad:
* **#1382** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1382?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pavellnk (approved):

devbydaniel (approved):

PR #1474, opened 2026-08-19
fix(letterheads): accept permission-encrypted letterhead PDFs (AYC-471)

sirolad:
Fixes [AYC-471](https://linear.app/ayunis/issue/AYC-471/briefpapier-pdf-upload-rejects-valid-file) — uploading a letterhead failed with "Bitte laden Sie eine gültige PDF-Datei hoch."

## Cause

`LetterheadPdfService` validated uploads with `PDFDocument.load(buffer)`, which **throws for any PDF carrying an encryption dictionary** — including files that are only locked against editing and open password-free in every viewer (that is why the upload dialog's pdf.js preview rendered the file while the backend refused it). The thrown parse error was swallowed and re-thrown as a single generic `LETTERHEAD_INVALID_PDF`, so neither the user nor the logs could tell an encrypted PDF from a corrupt or multi-page one.

`{ ignoreEncryption: true }` is not a fix: pdf-lib cannot decrypt streams, and the same library composites the letterhead behind exported documents. Verified locally — accepting an encrypted letterhead that way makes the export fail with `Unknown compression method in flate stream`.

## Change

- **Normalize on failure.** When pdf-lib cannot read an upload, it is rewritten through MuPDF (drops encryption, repairs broken cross-reference tables) and the normalized copy is stored. Uploads pdf-lib already accepts are stored byte-for-byte, so the common path is untouched and MuPDF's WASM runtime is only loaded on the fallback path. MuPDF documents and buffers are destroyed after use — their memory lives on the WASM heap, which JS garbage collection does not reclaim.
- **Say why an upload was refused.** New `LETTERHEAD_PDF_NOT_SINGLE_PAGE` and `LETTERHEAD_PDF_PASSWORD_PROTECTED` codes with German and English messages; `LETTERHEAD_INVALID_PDF` now carries the underlying parser error, and the previously swallowed failure is logged.
- PDFs needing a real user password are still rejected — decrypting those would require the password.

New dependency: `mupdf` (WASM, AGPL-3.0, same license as this repo). It is ESM-only with top-level await, so it is loaded via dynamic `import()`; the CommonJS build preserves the call (verified in `dist/`), and the compiled service was smoke-tested under Node 24.

## Verification

- New service specs cover: unchanged bytes for a plain PDF, permission-encrypted PDF (AES-256 and RC4-128) accepted and compositable, encrypted multi-page rejected, user-password PDF rejected, garbage rejected.
- Use-case spec asserts the **decrypted** copy is what gets uploaded to storage.
- Fixtures for the encrypted cases are generated in-test, and each asserts pdf-lib alone cannot read the input.
- Compiled service smoke-tested against PDFs encrypted by a third-party writer (pypdf, RC4-128 and AES-128): stored bytes load and composite in pdf-lib; 200 normalize cycles run without leaking or crashing.
- Full backend suite green (657 suites / 4588 tests) on Node 24; `nest build` clean.

## Not verified

The PDF from the ticket (`Testbrief – Stadt Sternfeld (1).pdf`) is behind Linear's authenticated upload URL, which this agent cannot fetch, so the fix is not confirmed against that exact file. Both plausible causes are covered: if it was permission-encrypted it now uploads, and if it was multi-page the user now gets "Das Briefpapier-PDF darf nur eine Seite enthalten." instead of the generic message. Re-testing with the original file in the demo account is the remaining step.

Linear Issue: [AYC-471](https://linear.app/ayunis/issue/AYC-471/briefpapier-pdf-upload-rejects-valid-file)

<div><a href="https://cursor.com/agents/bc-83e6d024-f423-4083-86a2-8564f217f7a2?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-83e6d024-f423-4083-86a2-8564f217f7a2&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/ayunis-core-backend/package.json
+++ b/ayunis-core-backend/package.json
@@ -98,6 +98,7 @@
     "minio": "^8.0.5",
     "mjml": "^4.7.1",
     "multer": "^2.2.0",
+    "mupdf": "^1.28.0",
     "nest-commander": "^3.19.1",
     "nestjs-cls": "^6.0.1",
     "nestjs-pino": "^4.6.1",
--- a/ayunis-core-backend/src/domain/letterheads/SUMMARY.md
+++ b/ayunis-core-backend/src/domain/letterheads/SUMMARY.md
@@ -23,7 +23,8 @@ letterheads/
 │   ├── ports/
 │   │   └── letterheads-repository.port.ts
 │   ├── services/
-│   │   └── letterhead-pdf.service.ts   # Shared PDF validation & storage path builder
+│   │   ├── letterhead-pdf.service.ts   # Shared PDF validation & storage path builder
+│   │   └── pdf-normalizer.service.ts   # MuPDF rewrite for PDFs pdf-lib cannot read
 │   └── use-cases/
 │       ├── create-letterhead/
 │       ├── delete-letterhead/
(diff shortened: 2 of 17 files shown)


sirolad:
Closing as superseded by #1634. QA with the original attachment confirmed it is an unencrypted two-page PDF; letterhead inputs intentionally require separate single-page first/continuation files.

PR #1519, opened 2026-08-25
fix(auth): keep logout lifecycle available with SSO disabled (AYC-367)

sirolad:
### Summary

- keep SSO entry routes disabled when `FEATURE_SSO_LOGIN_ENABLED=false`
- keep browser logout and signed OIDC back-channel logout available during rollback
- add a dedicated flag-off Playwright project covering the complete logout lifecycle
- bound the public logout token input and publish the limit in OpenAPI

### QA evidence

[Latest Browser E2E job](https://github.com/ayunis-core/ayunis-core/actions/runs/32859421976/job/97839960392) passed the dedicated flag-off project. The PR media comment contains desktop/mobile screenshots and logout GIFs from the same revision.

Fresh slot 2 database:

- all 153 migrations applied from an empty volume
- minimal seed completed
- `playwright test --project=sso-disabled`: 2/2 passed locally

The journey verifies:

- SSO discovery returns 404 with the feature disabled
- unauthenticated back-channel logout reaches request validation instead of the feature guard
- a password-authenticated user logs out through the account menu
- revisiting a protected route redirects to login, proving the session was revoked

### Validation

- backend: 715 suites / 4,831 tests passed
- focused backend: 2 suites / 16 tests passed
- backend typecheck and dependency checks passed
- frontend production build and typecheck passed
- E2E lint and typecheck passed
- pre-commit checks passed

Refs: AYC-367

PR #1520, opened 2026-08-25
docs(tooling): require feature flag on and off contracts (AYC-000)

sirolad:
## Summary

- require feature-gated changes to define enabled and disabled contracts
- replace the abstract \"test both states\" guidance with the concrete flag-off recipe that works here: second backend on a free port + dedicated Playwright project, modelled on the `sso-disabled` setup from #1519
- explain why mocking the guard/config is not proof: flags resolve from `process.env` at boot
- keep the detail in the `feature-toggles` skill; `.claude/CLAUDE.md` keeps the rule plus a pointer instead of a second copy
- fix the \"changing a default\" note to name where the deployed value actually lives (host `ayunis-core-backend/.env`)

## Validation

- restacked onto `main` @ 688f67e9d so the cited harness (`sso-disabled` project, `publicApi` fixture, `/api/health` poll) exists in-tree
- every cited path/endpoint verified present: `features.config.spec.ts`, `playwright.config.ts` `sso-disabled` project, `src/fixtures/test.ts` `publicApi`, `app.controller.ts` `@Get('health')`, `DEPLOYMENT.md` → Configuration Updates
- env precedence claim verified against `appsignal.cjs` (dotenv does not overwrite already-set vars, so shell env wins over `.env.dev` and `.env`)

---

> [!NOTE]
> **Low Risk**
> Documentation-only updates to agent guidelines and the feature-toggles skill; no runtime, auth, or deployment code changes.
> 
> **Overview**
> Codifies **two-state feature flag contracts** for agents: every gated change must define and verify behavior with the flag on and off, protect shared routes/workflows when new entry points are gated, and treat boot-time `process.env` resolution as the reason mocks are not enough for the disabled contract.
> 
> **`CLAUDE.md`** adds principle **§11** with the rule and a pointer to the skill so the long recipe is not duplicated.
> 
> **`feature-toggles` skill** expands with *define both contracts first*, a corrected *changing a default* note (host `ayunis-core-backend/.env`, CI e2e workflow), and a **Verification** section: config spec, flag-on e2e, when flag-off is required, and the concrete **flag-off e2e recipe** (second backend on a spare port, dedicated Playwright project, modeled on `sso-disabled`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8bdc1829f20a5356021bcb90c480d91fb3b5e45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/skills/feature-toggles/SKILL.md
+++ b/.claude/skills/feature-toggles/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: feature-toggles
-description: Add, modify, or remove feature toggles. Use when gating new features or changing toggle defaults.
+description: Add, modify, remove, or review feature toggles. Use when gating a feature, changing a toggle default, or changing behavior shared with a feature-gated path.
 ---
 
 # Feature Toggles
@@ -12,6 +12,14 @@ description: Add, modify, or remove feature toggles. Use when gating new feature
 - Endpoint: `ayunis-core-backend/src/app/presenters/http/app.controller.ts` — `GET /feature-toggles`
 - Response DTO: `ayunis-core-backend/src/app/presenters/http/dto/feature-toggles-response.dto.ts`
 - Frontend hooks: `ayunis-core-frontend/src/features/feature-toggles/`
+
+## Define both contracts first
+
+Before implementation, state the observable behavior with the flag enabled and disabled. Identify any existing workflow that shares a route, controller, navigation action, service, or persisted state with the gated feature.
+
+- Gate the narrowest new entry point that satisfies the product contract.
+- Keep shared cleanup, revocation, logout, recovery, and existing-user paths available unless the disabled-state contract explicitly removes them.
+- Do not switch an existing shared consumer to a gated endpoint without proving its flag-off behavior.
 
 ## Adding a new toggle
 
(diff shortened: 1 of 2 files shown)


sirolad:
* **#1520** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1520?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1520?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.claude/skills/feature-toggles/SKILL.md
```diff
@@ -33,7 +41,27 @@ description: Add, modify, or remove feature toggles. Use when gating new feature
...
+
+- `.github/workflows/e2e-tests.yml` — steps `Start backend with SSO login disabled` and `Test logout lifecycle with SSO login disabled`: the same build artifact started again with the flag off on a spare port, health-polled, log added to the `backend-log` artifact. Same job as the flag-on run: one extra process and one spec, not a matrix.
+- `ayunis-core-e2e/playwright.config.ts` — the `sso-disabled` project. The spec must also be listed in the `chromium` project's `testIgnore`, or it runs in both states.
+- `ayunis-core-e2e/tests/auth/sso-disabled-logout.spec.ts` — prove the feature is gone first (gated route 404s; `publicApi` fixture for unauthenticated endpoints), then exercise the shared workflow that must stay available.
+
+Locally: `FEATURE_<NAME>_ENABLED=false pnpm run start:dev` in a second slot. Shell env wins over `.env.dev` and `.env`; never hand-edit the generated `.env.dev`.
+
+Mocked guards, mocked API clients, and direct controller invocation exercise wiring, not runtime composition: they do not satisfy the OFF contract for a shared path.
+
+Record both ON and OFF results in the PR validation summary.
```
devbydaniel, line 64: Not sure if we need this to be honest. That adds another layer of complexity to our already complex end to end and verification setup.

sirolad: You have a valid point. But existing feature flags will not have e2e retroactively. Also, feature flags would eventually be cleaned up, this is just to ensure that our tests cover both state when they are active. I think we should keep this.

pavellnk: @devbydaniel, i’m with Suraj on this. My point is that if we implement something behind a feature flag, it’s probably a significant change, and five minutes more or less won’t matter much.

pavellnk (approved):

PR #1574, opened 2026-08-31
feat(sso): persist local password login policy (AYC-868)

sirolad:
## Summary

- persist the per-organization local-password-login policy on SSO connections
- default existing and new organizations to local password access
- add the repository locking support required for atomic policy changes

## QA

- migration up/down and repository integration tests
- backend targeted suite: 25 suites, 166 tests passed across the complete stack

Linear: AYC-868

---

> [!NOTE]
> **Medium Risk**
> Touches authentication-related SSO connection data and adds policy constraints; enforcement at login is not in this diff, but incorrect defaults or locking could affect future password session flows.
> 
> **Overview**
> Adds a per-org **`localPasswordLoginEnabled`** flag on SSO connections so tenants can allow password login alongside SSO or enforce SSO-only, while **defaulting to `true`** for new and migrated rows.
> 
> A DB migration and entity/schema/mapper updates persist the field, with a **CHECK** that at least one of local password login or SSO (`enabled`) stays on. The repository gains **narrow reads** of the flag (plain vs **pessimistic read lock** for session issuance) and a **compare-and-swap** update that toggles local password login without changing SSO enablement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe349c690ab97c8b9ca178a0c40e1782ae43d13e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sirolad:
* **#1601** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1601?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1577** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1577?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1576** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1576?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1575** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1574** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1574?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pavellnk (approved):

ayunis-core-backend/src/iam/sso/infrastructure/persistence/postgres/org-sso-connections.repository.ts
```diff
@@ -57,6 +57,35 @@ export class PostgresOrgSsoConnectionsRepository extends OrgSsoConnectionsReposi
...
+    };
+    const record = lockForSessionIssuance
+      ? await this.records.findOne({
+          ...options,
+          lock: { mode: 'pessimistic_read' },
+        })
+      : await this.records.findOne(options);
+    return record?.localPasswordLoginEnabled ?? null;
+  }
+
```
devbydaniel, line 88: This feels a little over engineered to be honest.

- What is the scenario in which we need a lock? I this just to avoid a race condition in which an admin sets the attribute in EXACTLY the same time as someone tries to log in? If this is the case I'd say we don't need it. Or is there another scenario?
- If we really need the option to lock, I would let the caller of the repository decide if they want the lock and not introduce two separate repository methods. In other words, I'd add a second `withLock: boolean` property or something like that to the port method

ayunis-core-backend/src/iam/sso/infrastructure/persistence/postgres/schema/org-sso-connection.record.ts
```diff
@@ -22,6 +22,7 @@ import { OrgSsoEmailDomainRecord } from 'src/iam/sso/infrastructure/persistence/
   '"zitadelOrgId" IS NULL OR ("zitadelOrgId" <> \'\' AND "zitadelOrgId" = btrim("zitadelOrgId"))',
 )
 @Check('NOT "enabled" OR "zitadelOrgId" IS NOT NULL')
+@Check('"localPasswordLoginEnabled" OR "enabled"')
```
devbydaniel, line 25: can you explain this? `localPasswordLoginEnabled` must be `true` if `enabled` is `false`?

this feels a little counter intuitive to me. `localPasswordLoginEnabled` to me is a property in the context of the SSO connection. If the SSO connection is disabled, I wouldn't even look at `localPasswordLoginEnabled` because the connection itself is disabled so whatever it has stored inside it does not count.

ayunis-core-backend/src/iam/sso/domain/org-sso-connection.entity.ts
```diff
@@ -17,6 +17,7 @@ export interface OrgSsoConnectionParams {
   zitadelIdpId?: string | null;
   enabled?: boolean;
   jitProvisioningEnabled?: boolean;
+  localPasswordLoginEnabled?: boolean;
```
devbydaniel, line 20: Just a minor point, I'd find `enforced` more descriptive of what we mean here

devbydaniel (changes requested):

ayunis-core-backend/src/iam/sso/infrastructure/persistence/postgres/org-sso-connections.repository.ts
sirolad: The lock is for that race, but the window is wider than an exact instant. Without it, a password login can read “allowed”, SSO-only activation can then update the policy and revoke every existing password session, and the login can insert a new password session after the revoke. The shared lock makes the result deterministic: either issuance commits first and activation revokes it, or activation commits first and issuance sees the disabled policy. I would keep the two repository methods because they describe different intent. The caller already chooses whether it needs session-issuance serialization; a generic withLock boolean would expose a persistence mechanism rather than the operation we need.

ayunis-core-backend/src/iam/sso/infrastructure/persistence/postgres/schema/org-sso-connection.record.ts
sirolad: Yes, that is intentional: when SSO is disabled, local password login must be enabled. The flag is the organization’s authentication policy stored alongside its SSO connection, rather than a dormant option that we ignore when the connection is disabled. The constraint prevents an organization from ending up with both login methods disabled. It also avoids retaining a hidden SSO-only setting that would unexpectedly take effect when SSO is re-enabled. The application therefore requires local login to be restored before SSO can be disabled.

ayunis-core-backend/src/iam/sso/domain/org-sso-connection.entity.ts
sirolad: I would keep localPasswordLoginEnabled here. enforced is ambiguous about what is being enforced and would invert the meaning of the current flag. This name states the capability directly and stays consistent across the entity, persistence model, API, and policy checks.

ayunis-core-backend/src/iam/sso/infrastructure/persistence/postgres/org-sso-connections.repository.ts
```diff
@@ -57,6 +57,35 @@ export class PostgresOrgSsoConnectionsRepository extends OrgSsoConnectionsReposi
...
+      where: { orgId },
+      select: { localPasswordLoginEnabled: true },
+    };
+    const record = lockForSessionIssuance
+      ? await this.records.findOne({
+          ...options,
+          lock: { mode: 'pessimistic_read' },
+        })
+      : await this.records.findOne(options);
+    return record?.localPasswordLoginEnabled ?? null;
```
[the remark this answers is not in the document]
sirolad: No code change here. The claim does not reproduce with TypeORM or the database-backed flow: a partial findOne select returns the requested column without requiring the primary key. The stack E2E persists localPasswordLoginEnabled=false, reads false through discovery, and rejects direct password login. Both assertions would fail open if this query returned null.

ayunis-core-backend/src/iam/sso/domain/org-sso-connection.entity.ts
```diff
@@ -53,6 +55,7 @@ export class OrgSsoConnection {
         : normalizeZitadelIdpId(params.zitadelIdpId);
     this.enabled = params.enabled ?? false;
     this.jitProvisioningEnabled = params.jitProvisioningEnabled ?? false;
+    this.localPasswordLoginEnabled = params.localPasswordLoginEnabled ?? true;
```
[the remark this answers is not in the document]
sirolad: No code change here. This state is excluded by the persistence and application invariants. Configuration updates are allowed only while SSO is disabled, and the database check localPasswordLoginEnabled OR enabled requires every disabled row to have local password login enabled. An SSO-only connection must restore local password login before SSO can be disabled, while enabled connections are rejected before updatedConnection is reached.

devbydaniel (approved):

sirolad:
### Merge activity

* **Sep 14, 4:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1574).
* **Sep 14, 4:31 PM UTC**: @sirolad merged this pull request with [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1574).

PR #1575, opened 2026-08-31
feat(auth): enforce organization SSO-only sessions (AYC-868)

sirolad:
## Summary

- enforce organization policy for password login, MFA completion and refresh
- serialize session issuance with SSO-only activation and revoke password-origin refresh sessions
- commit rejected MFA attempts before returning the error, using one database connection; roll back valid factor consumption if session issuance fails

## Release constraint

Refresh revocation is immediate. Existing stateless access tokens remain valid until expiry; keep JWT_EXPIRES_IN at or below the accepted one-hour ceiling.

## QA

- local slot 10: 132 backend suites / 664 tests across the complete stack
- real PostgreSQL: password and SSO MFA lockout, concurrent rejections with a two-connection pool, and successful-factor rollback
- 15 browser journeys passed across existing auth and Superadmin SSO tests (plus setup)
- staged lint, typecheck, complexity and dependency checks passed

Linear: AYC-868

---

> [!NOTE]
> **High Risk**
> Changes authentication, MFA, refresh rotation, and org policy under transactional locking; mis-coordination could block logins or leave stale password sessions, though existing JWT access tokens still expire naturally.
> 
> **Overview**
> Adds **organization SSO-only login**: when local password login is disabled, password login, MFA continuation, and password-origin refresh return **403** with a clear SSO message instead of issuing sessions.
> 
> **`LocalPasswordLoginPolicyService`** reads org policy from SSO (SSO flows skip the check). Password **login**, **refresh**, and **`CompleteMfaLoginUseCase`** run in **transactions** that **lock the org policy row** before creating or rotating sessions, so concurrent SSO-only activation and session issuance cannot race. **`LoginUseCase`** now builds access-token claims from the user returned by **`AuthorizeUserLoginUseCase`** inside that transaction.
> 
> MFA completion is centralized in **`CompleteMfaLoginUseCase`**: policy lock → verify TOTP or confirm enrollment → session creation in one transaction, so a failed issuance **rolls back** consumed factors; invalid codes still **commit** failed-attempt accounting before the error is returned. **`MfaLoginController`** uses this boundary and clears the pending cookie on terminal policy/account errors.
> 
> **Sessions**: **`PrepareSessionRotationUseCase`** validates opaque refresh tokens before rotation so refresh can enforce policy first; **`RevokePasswordSessionsForOrgUseCase`** bulk-revokes password sessions when SSO-only is turned on. MFA and session repos participate in **ambient transactions** via **`TransactionHost`**.
> 
> **SSO**: new **`SsoConnectionPolicyModule`** with **`GetOrgAuthenticationPolicyUseCase`** / **`SetOrgLocalPasswordLoginEnabledUseCase`** (reviewed mapping confirmation, revokes password sessions on disable). Shared **`ReviewedSsoMapping`** for enable/password-policy mutations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c987a318f1f39d55a3a014ee80c97b577e13463. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sirolad:
* **#1601** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1601?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1577** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1577?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1576** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1576?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1575** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1574** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1574?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ayunis-core-backend/src/iam/sso/sso-connection-policy.module.ts
```diff
@@ -0,0 +1,32 @@
...
+      useClass: PostgresOrgSsoConnectionsRepository,
+    },
+    GetOrgAuthenticationPolicyUseCase,
+    SetOrgLocalPasswordLoginEnabledUseCase,
+  ],
+  exports: [
+    OrgSsoConnectionsRepository,
+    GetOrgAuthenticationPolicyUseCase,
+    SetOrgLocalPasswordLoginEnabledUseCase,
+  ],
```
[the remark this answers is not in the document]
sirolad: No code change is needed here. SsoConnectionPolicyModule is an internal Nest composition module inside the same iam/sso bounded module. Authentication, Users, and Invites consume only its exported policy use cases; the repository export is consumed solely by the remaining SSO use cases in the same bounded module. The SSO module documentation now states this boundary explicitly.

ayunis-core-backend/src/iam/sso/application/sso.errors.ts
```diff
@@ -82,6 +83,17 @@ export class SsoConnectionMustBeDisabledError extends SsoError {
...
+export class SsoMustRemainEnabledError extends SsoError {
+  constructor(orgId: UUID) {
+    super(
+      `SSO must remain enabled for organization '${orgId}' while local password login is disabled`,
+      SsoErrorCode.MUST_REMAIN_ENABLED,
+      409,
+      { orgId },
+    );
+  }
+}
```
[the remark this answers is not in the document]
sirolad: This guard is implemented in the immediate up-stack policy-management PR 1576, where disabling local password login is introduced. PR 1575 has no application path that can set localPasswordLoginEnabled to false: the persisted value defaults to true and the policy mutation does not exist until PR 1576. The stacked state therefore never exposes the constraint as an unexpected 500; PR 1576 adds the explicit SsoMustRemainEnabledError guard and its regression test together with the mutation capability.

ayunis-core-backend/src/iam/authentication/presenters/http/mfa-login.controller.ts
```diff
@@ -150,6 +159,27 @@ export class MfaLoginController {
     return this.mfaPendingJwtService.verify(token);
   }
 
+  private async assertPendingLoginAllowed(
```
pavellnk: @sirolad, this is application logic. Let’s keep the controllers small, tiny, clear, and sexy.

ayunis-core-backend/src/iam/sso/application/reviewed-sso-mapping.ts
```diff
@@ -0,0 +1,42 @@
+import { InvalidSsoConfigurationError } from 'src/iam/sso/application/sso.errors';
```
pavellnk: @sirolad, this file feels a bit out of place. Why is it directly in the application folder? Please find a more appropriate location that follows the project conventions.

pavellnk (approved):
Looks great! Just a few tiny things regarding placement. Please check them, and then feel free to merge

ayunis-core-backend/src/iam/authentication/presenters/http/mfa-login.controller.ts
sirolad: Fixed. The MFA controller now delegates the pending-login policy decision and user lookup to MfaPendingLoginPolicyService; the controller only handles HTTP state and completes the login. The service and controller paths are covered by focused tests.

ayunis-core-backend/src/iam/sso/application/reviewed-sso-mapping.ts
sirolad: Fixed. ReviewedSsoMapping now lives under application/models as an application model with its matching behavior encapsulated, instead of sitting directly under application.

ayunis-core-backend/src/iam/authentication/application/use-cases/login/login.use-case.ts
```diff
@@ -35,17 +39,20 @@ export class LoginUseCase {
...
         command.user.id,
         command.authenticationMethod,
         command.zitadelSessionId,
       ),
     );
-
+    await this.localPasswordLoginPolicy.assertSessionIssuanceAllowed(
+      command.user.orgId,
+      command.authenticationMethod,
+    );
```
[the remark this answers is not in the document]
sirolad: Valid and fixed in the immediate up-stack PR 1576, where the SSO-only cutover mutation is introduced. The cutover now revokes password sessions before the guarded policy update, so refresh and cutover acquire locks in session-to-policy order; both operations remain atomic in the same transaction. A regression test asserts this lock order. PR 1575 alone has no path that performs the cutover.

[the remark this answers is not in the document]
sirolad: Follow-up correction: the revoke-first approach in my previous reply exposed a session-gap race and has been replaced. The final design uses policy → session ordering everywhere: login and refresh acquire the shared policy lock before creating or rotating a password session, while cutover updates the policy under its exclusive lock before revoking password sessions. Opaque refresh now separates validation from rotation so it can acquire the policy lock before the write. This removes the opposing lock order without creating a post-revoke issuance window.

ayunis-core-backend/src/iam/sessions/infrastructure/repositories/local/local-refresh-tokens.repository.ts
```diff
@@ -34,7 +40,7 @@ export class LocalRefreshTokensRepository extends RefreshTokensRepository {
     successor: RefreshToken,
   ): Promise<boolean> {
     try {
-      await this.repository.manager.transaction(async (manager) => {
+      await this.txHost.tx.transaction(async (manager) => {
```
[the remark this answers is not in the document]
sirolad: Not reproducible; no code change is needed. Inside @Transactional(), txHost.tx is the transaction-scoped TypeORM EntityManager. TypeORM 0.3.30 reuses that manager\u0027s QueryRunner, and PostgresQueryRunner starts a SAVEPOINT when transactionDepth is already greater than zero. I verified this against the existing slot-4 PostgreSQL instance: the live probe reported outer depth 1, nested depth 2, and depth 1 after the nested call, with no error. The savepoint is deliberate here: a lost conditional rotation must roll back the just-inserted successor while returning false and preserving the outer refresh transaction and policy lock.

ayunis-core-backend/src/iam/sso/sso-connection-policy.module.ts
```diff
@@ -0,0 +1,32 @@
...
+})
+export class SsoConnectionPolicyModule {}
```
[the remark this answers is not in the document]
sirolad: Fixed in the latest revision. SSO SUMMARY now documents SsoConnectionPolicyModule, the two policy use cases, ReviewedSsoMapping, the localPasswordLoginEnabled default, and the internal repository boundary.

ayunis-core-backend/src/iam/authentication/application/services/local-password-login-policy.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    authenticationMethod: SessionAuthenticationMethod,
+  ): Promise<void> {
+    if (authenticationMethod === SessionAuthenticationMethod.SSO) return;
+    const policy = await this.getOrgAuthenticationPolicy.execute(
+      new GetOrgAuthenticationPolicyQuery(orgId),
+    );
+    if (!policy.localPasswordLoginEnabled) {
+      throw new LocalPasswordLoginDisabledError();
+    }
+  }
```
devbydaniel: I don't get this. If the auth method is SSO, local password are always allowed. If the auth method is NOT SSO, we check an SSO use case if the SSO config contains the flag for local passwords allowed – even though the auth method is not SSO?

My understanding would be exactly the opposite:

- If the auth method is not SSO => return
- If the auth method is SSO => check the SSO connection if it contains the `passwordAllowed` flag

ayunis-core-backend/src/iam/authentication/application/services/local-password-login-policy.service.ts
```diff
@@ -0,0 +1,39 @@
...
+  ): Promise<void> {
+    if (authenticationMethod === SessionAuthenticationMethod.SSO) return;
+    const policy = await this.getOrgAuthenticationPolicy.execute(
+      new GetOrgAuthenticationPolicyQuery(orgId, true),
+    );
+    if (!policy.localPasswordLoginEnabled) {
+      throw new LocalPasswordLoginDisabledError();
+    }
+  }
+}
```
devbydaniel, line 52: this is exactly the same method as above, why do we need this here?

ayunis-core-backend/src/iam/authentication/application/services/mfa-pending-login-policy.service.ts
devbydaniel, line 1: Why do we need a separate service for this, tied to MFA? This checks if local passwords are allowed in the MFA flow – so why not add a `assertAllowedForUser` vs `assertAllowedForOrg` to the `LocalPasswordPolicyService` – this way the assertion is not tied to MFA but simply tells you if that user is allowed to authenticate via local passwords

ayunis-core-backend/src/iam/authentication/application/use-cases/refresh-token/refresh-token.use-case.ts
```diff
@@ -71,18 +82,21 @@ export class RefreshTokenUseCase {
    * FUTURE(AYC-452): remove ~7 days after deploy, once legacy JWT refresh
    * tokens have all expired.
    */
```
devbydaniel, line 83: <p class='graphite__hidden'><i>[Re: lines +78 to +84]</i></p>
Why do we need this? Can we just reject stale tokens so that users need to log in again without a `legacy` ceremony?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575?utm_source=unchanged-line-comment">Graphite</a>.</p>

ayunis-core-backend/src/iam/sessions/application/ports/refresh-tokens.repository.ts
```diff
@@ -35,6 +35,8 @@ export abstract class RefreshTokensRepository {
 
   abstract revokeSsoForUser(userId: UUID): Promise<void>;
 
+  abstract revokePasswordForOrg(orgId: UUID): Promise<void>;
```
devbydaniel: This should mean `revokePasswordsForOrg` (passwords plural), right?

ayunis-core-backend/src/iam/sessions/application/models/prepared-session-rotation.ts
devbydaniel, line 1: This feels redundant, we simply access properties of the refresh token. Is there any reason for this?

devbydaniel (changes requested):

ayunis-core-backend/src/iam/authentication/application/services/local-password-login-policy.service.ts
sirolad: SSO is the bypass here intentionally. This service answers whether the supplied authentication method is allowed by the local-password policy. An SSO session did not use a local password, while PASSWORD must be checked against the organization policy. Reversing this would allow password sessions and reject SSO sessions when SSO-only mode is active, which is the opposite of the requirement.

ayunis-core-backend/src/iam/authentication/application/use-cases/refresh-token/refresh-token.use-case.ts
sirolad: Agreed that the migration path is stale, but I am keeping its removal separate from this SSO-only change so we can verify the deployed refresh-token TTL before deleting compatibility code. I created AYC-895 for that cleanup: https://linear.app/ayunis/issue/AYC-895/remove-expired-legacy-jwt-refresh-token-migration-path

ayunis-core-backend/src/iam/authentication/application/services/local-password-login-policy.service.ts
sirolad: These methods enforce the same rule but have different consistency contracts. assertAllowedForOrg is a normal read used before issuing an MFA-pending token. assertSessionIssuanceAllowed takes the shared policy-row lock inside the create/rotate transaction and holds it through the session write, serializing issuance with SSO-only activation. I kept the intent-specific operations instead of exposing locking as a caller-controlled boolean.

ayunis-core-backend/src/iam/sessions/application/models/prepared-session-rotation.ts
sirolad: Fixed. PrepareSessionRotationUseCase now returns the validated RefreshToken directly and RotateSessionCommand accepts it directly. The two-stage validate → policy lock → atomic rotate ordering is unchanged.

ayunis-core-backend/src/iam/authentication/application/services/mfa-pending-login-policy.service.ts
sirolad: Fixed. I folded the user lookup into LocalPasswordLoginPolicyService.assertAllowedForUser and removed MfaPendingLoginPolicyService. The MFA controller still owns cookie cleanup, and the user is fetched only once and returned for session completion.

ayunis-core-backend/src/iam/sessions/application/ports/refresh-tokens.repository.ts
sirolad: Fixed. The repository operation is now named revokePasswordSessionsForOrg, matching the use case and making the affected resource explicit.

ayunis-core-backend/src/iam/sso/SUMMARY.md
```diff
@@ -21,7 +21,9 @@ Logout is backend-owned. The feature-independent `LogoutController` serves `auth
...
 
-`OrgSsoConnection` owns normalized connection state, while `OrgSsoConnectionsRepository` defines lookups, persistence, and conditional updates for routing and runtime flags. The Postgres adapter reports database constraint violations without deciding application conflicts and uses conditional writes to prevent stale operator changes from overwriting newer state.
+`SsoConnectionPolicyModule` owns connection persistence and exports the authentication-policy use cases consumed by Authentication, Users, and Invites. `GetOrgAuthenticationPolicyUseCase` reads whether local password login is allowed; `SetOrgLocalPasswordLoginEnabledUseCase` changes that policy and revokes password-origin sessions when SSO-only is enforced. `ReviewedSsoMapping` captures the operator-reviewed domains and broker identifiers so a concurrent remap cannot activate an unseen policy. The module is an internal Nest composition boundary within SSO; its repository port is shared only with other SSO use cases.
```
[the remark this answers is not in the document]
sirolad: No code change here. This sentence describes the completed stack boundary. PR 1575 first exposes the policy module to Authentication; the immediate up-stack PR 1576 moves SsoModule onto it and adds Users and Invites as consumers. The final stacked state matches the SUMMARY and avoids duplicating a transient description that the next commit would immediately rewrite.

ayunis-core-backend/src/iam/sso/application/use-cases/set-org-sso-enabled/set-org-sso-enabled.use-case.ts
```diff
@@ -34,9 +33,11 @@ export class SetOrgSsoEnabledUseCase {
...
     }
     const existing = existingState.connection;
-    if (!this.matchesReviewedMapping(existing, command.reviewedMapping)) {
-      throw new SsoConnectionChangedError(command.orgId);
-    }
+    this.assertReviewedMappingMatches(
+      existing,
+      command.reviewedMapping,
+      command.orgId,
+    );
```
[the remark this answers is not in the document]
sirolad: No code change in this PR. PR 1575 introduces enforcement but has no operation that can set localPasswordLoginEnabled to false; the persisted and domain defaults are true. The immediate up-stack PR 1576 introduces the mutation and guard together: assertCanSetSsoState throws SsoMustRemainEnabledError before setEnabled, and its regression test covers this case.

ayunis-core-backend/src/iam/authentication/application/services/local-password-login-policy.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    private readonly getOrgAuthenticationPolicy: GetOrgAuthenticationPolicyUseCase,
+    private readonly findUserById: FindUserByIdUseCase,
+  ) {}
+
+  async assertAllowedForOrg(
+    orgId: UUID,
+    authenticationMethod: SessionAuthenticationMethod,
+  ): Promise<void> {
+    if (authenticationMethod === SessionAuthenticationMethod.SSO) return;
+    const policy = await this.getOrgAuthenticationPolicy.execute(
```
devbydaniel, line 23: convert this use case into `GetSSOConnection`, then

```suggestion

    const ssoConnection = await this.getSSOConnectionUseCase.execute()
      .catch((e) => if e instanceof NotFoundError return else throw e))
    if (!ssoConnection.localPasswordLoginEnabled) throw DisabledError()
```

ayunis-core-backend/src/iam/sso/application/use-cases/get-org-authentication-policy/get-org-authentication-policy.use-case.ts
```diff
@@ -0,0 +1,31 @@
...
+import { OrgSsoConnectionsRepository } from 'src/iam/sso/application/ports/org-sso-connections.repository';
+import { UnexpectedSsoError } from 'src/iam/sso/application/sso.errors';
+import { GetOrgAuthenticationPolicyQuery } from 'src/iam/sso/application/use-cases/get-org-authentication-policy/get-org-authentication-policy.query';
+
+export interface OrgAuthenticationPolicy {
+  localPasswordLoginEnabled: boolean;
+}
+
+@Injectable()
+export class GetOrgAuthenticationPolicyUseCase {
```
devbydaniel, line 12: replace with `GetSSOConnectionUseCase` and let the caller decide what to do with it. In case of the `OrgAuthenticationPolicy`, that policy will then read the correct property and make the final decision about the org as a whole (see code example in the policy)

devbydaniel (approved):

sirolad:
### Merge activity

* **Sep 14, 4:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575).
* **Sep 14, 4:32 PM UTC**: [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575) rebased this pull request as part of a merge.
* **Sep 14, 4:32 PM UTC**: @sirolad merged this pull request with [Graphite](https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1575).

PR #1583, opened 2026-09-01
fix(super-admin): guard every super admin route at the layout (AYC-000)

sirolad:
## Problem

Seven of eleven super-admin routes carried no role guard. A regular user opening one by URL ran the route loader, hit the backend 403, and landed on the generic error screen — whose only affordance is a reload that fails identically.

| Route | Guard before | Behaviour for a non-super-admin |
| --- | --- | --- |
| `/super-admin-settings` | ✅ | redirect `/` |
| `academy`, `anonymization`, `app-alerts` | ✅ | redirect `/` |
| `users`, `orgs`, `orgs/$id`, `skills`, `super-admins`, `models-catalog` | ❌ | loader 403 → "Ein unerwarteter Fehler ist aufgetreten" |
| `platform-config` | ❌ | no loader either — rendered a broken super-admin page shell |

The backend was always correct: `SystemRolesGuard` returns 403 and audit-logs the denial. The sidebar entry is also correctly gated, so the way in is a direct URL — a bookmark, a shared link, or a demoted super-admin.

Root cause: the guard was copy-pasted per route instead of living on a shared parent, so new routes silently shipped without it.

## Fix

Adds `super-admin-settings.tsx` — a layout route mirroring the existing `admin-settings.tsx` pattern: `Outlet` plus one `beforeLoad` role check. All eleven routes now hang off it, and the four duplicated guards are gone. `beforeLoad` runs before any child loader, so the 403 never happens.

No new 403 page: `admin-settings` redirects home rather than explaining, and a user who cannot see the sidebar link has no reason to be told the section exists.

## Verification

- New e2e spec `tests/super-admin/access-control.spec.ts` — org admin (non-super-admin) hits all eleven routes, asserts redirect to `/chat`.
- **Confirmed the spec fails without the fix.** With the layout route removed, the failure artifact captures the reported bug verbatim: heading `Ein Fehler ist aufgetreten`, body `Ein unerwarteter Fehler ist aufgetreten. Bitte laden Sie die Seite neu, um fortzufahren.`
- Allow-path regression covered by the existing `users-list.spec.ts` and `account-lockout.spec.ts`, which log in as a real super admin — both still pass.
- `vite build`, `tsc --noEmit`, `eslint --max-warnings=0`, e2e lint + typecheck all clean.

```
✓ tests/super-admin/access-control.spec.ts  — redirected off every super admin route
✓ tests/super-admin/users-list.spec.ts      — super admin searches users and opens their organization
✓ tests/super-admin/account-lockout.spec.ts — super admin sees and unlocks a locked account
4 passed
```

## Before / after

Same principal both sides: an org admin **without** the `super_admin` system role, opening `/super-admin-settings/users` directly. Captured locally against a seeded e2e stack; the "before" column is this branch with the new layout route removed.

| Before | After |
| --- | --- |
| before desktop | after desktop |

<details><summary>Mobile (375px)</summary>

| Before | After |
| --- | --- |
| before mobile | after mobile |

</details>

The automated PR-media comment below adds the desktop/mobile GIF of the `platform-config` redirect.

### DIFF
--- a/ayunis-core-e2e/tests/super-admin/access-control.spec.ts
+++ b/ayunis-core-e2e/tests/super-admin/access-control.spec.ts
@@ -1,0 +1,26 @@
+import { test, expect } from '../../src/fixtures/test';
+
+// Every super-admin route, including the ones that used to carry no route
+// guard and surfaced the backend 403 as the generic error screen.
+const SUPER_ADMIN_ROUTES = [
+  '/super-admin-settings',
+  '/super-admin-settings/orgs',
+  '/super-admin-settings/orgs/00000000-0000-0000-0000-000000000000',
+  '/super-admin-settings/users',
+  '/super-admin-settings/skills',
+  '/super-admin-settings/super-admins',
+  '/super-admin-settings/models-catalog',
+  '/super-admin-settings/platform-config',
+  '/super-admin-settings/academy',
+  '/super-admin-settings/anonymization',
+  '/super-admin-settings/app-alerts',
+];
+
+test('org admin without the super admin system role is redirected off every super admin route', async ({
+  page,
+}) => {
+  for (const route of SUPER_ADMIN_ROUTES) {
+    await page.goto(route);
+    await expect(page).toHaveURL(/\/chat/);
+  }
+});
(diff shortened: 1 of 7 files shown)


sirolad:
* **#1583** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1583?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pavellnk (approved):
well done, nice finding!

PR #1586, opened 2026-09-02
fix(dev): terminate the full process tree on ./dev down (AYC-877)

sirolad:
Fixes AYC-877.

## Problem

`./dev down` printed `Slot N stopped.` while the backend and frontend node
processes stayed alive and listening, so the next `./dev up --slot N` failed with
`port 3080 is occupied by PID … not managed by this checkout`.

Two independent defects:

1. `_terminate_process_tree` used `pkill -TERM -P "$pid"`, which is single-level.
   The launch chain is three to four levels deep (`infisical run` → pnpm shim →
   pnpm → node), so the leaf that owns the socket never received a signal. The
   process-group kill alongside it only helps when `$pid` is the group leader,
   which is not reliable — in the non-tmux `nohup … &` branch the background job
   inherits the `dev` script's own process group, so `kill -- -$pid` fails with
   ESRCH.
2. `cmd_down` never checked whether the ports actually freed, so the failure was
   silent until the next `up`.

## Fix

- `_descendant_pids` walks the full descendant tree via `pgrep -P`, emitting
  deepest-first so a supervising parent (`nest --watch`) cannot respawn a leaf
  after it was signalled. The process-group kill is kept: a descendant already
  reparented away is invisible to the `pgrep` walk but still in the group.
- `cmd_down` verifies both slot ports with the existing `_listener_pids` helper
  and exits non-zero naming the holding PIDs instead of claiming success.

Chose the descendant walk over `setsid` because `setsid` is not available on
macOS by default, which is the primary platform for this script.

## Verification

`bash scripts/dev-status.test.sh` — full suite passes on the fix and both new
assertions fail on unmodified `dev`, reproducing the ticket:

```
# against HEAD's dev (before the fix):
Expected dev down to fail loudly while a slot port is still held.
Expected dev down to terminate the grandchild holding the port (PID 49295 survived).
exit 1

# against the fixed dev:
dev script tests passed
exit 0
```

New coverage:

- A real three-level process chain (`bash` → `bash` → `sleep`) with only `docker`
  and `lsof` faked, so the termination runs against real `pgrep`/`ps`/`kill`. It
  asserts the grandchild — the position the port-owning node process occupies — is
  dead after `./dev down`.
- `dev down` exits non-zero and reports `Backend port 3970 (PID 999997)` while a
  listener still holds a slot port.

Ran on macOS (darwin 25.6.0), local worktree, no dev stack booted.

### Not verified

I could not reproduce the port-level symptom (`lsof … -sTCP:LISTEN` after `down`)
in this environment: a listener spawned from the agent's process tree shows up in
`lsof` as `(CLOSED)` rather than `LISTEN`, so `_listener_pids` cannot observe it.
The tests above cover the causal mechanism instead — the leaf surviving the
signal is exactly why the socket stayed held.

## Known limit (pre-existing, not addressed)

If a pidfile holds a stale PID that the OS has since reused, teardown now kills
that process's whole subtree rather than just it and its direct children. Adding
an ownership check to `_kill_pid` would change `up`'s restart path too, so I left
it alone rather than growing the fix.

## Follow-up

`scripts/dev-status.test.sh` runs in no CI workflow or hook — only manually. I
drafted wiring it into `deploy-script-tests.yml` but dropped it: I can't verify
that suite passes on `ubuntu-latest` from here, and shipping a possibly-red job
isn't worth it in this PR. Worth a separate ticket.

### DIFF
--- a/dev
+++ b/dev
@@ -433,6 +433,20 @@ cmd_down() {
   # Stop Docker infrastructure
   dc down
 
+  local entry label port pids leftovers=""
+  for entry in "Backend:$BACKEND_PORT" "Frontend:$FRONTEND_PORT"; do
+    label="${entry%%:*}"
+    port="${entry##*:}"
+    pids="$(_listener_pids "$port")"
+    if [ -n "$pids" ]; then
+      leftovers="$leftovers $label port $port (PID $pids);"
+    fi
+  done
+
+  if [ -n "$leftovers" ]; then
+    die "Slot $SLOT is still occupied —$leftovers stop those processes or use another slot."
+  fi
+
   info "Slot $SLOT stopped."
 }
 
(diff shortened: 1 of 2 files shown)


sirolad:
* **#1586** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1586?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1586?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1598, opened 2026-09-03
fix(email): block MJML template injection (AYC-890)

sirolad:
> [!NOTE]
> **Medium Risk**
> Security-sensitive email rendering changes; behavior is narrowed (no MJML includes) and escaping is added, with limited blast radius if options are applied consistently across templates.
> 
> **Overview**
> Hardens MJML email rendering against **template injection** when user-controlled strings are interpolated into markup.
> 
> All templates now compile MJML through a shared **`renderMjml`** helper that calls `mjml2html` with **`ignoreIncludes: true`**, so injected `<mj-include>` tags cannot pull in arbitrary files. **`escapeAttribute`** is exported (replacing the private `escapeAttr`) and used for banner alt text; the **email confirmation** template additionally escapes `confirmationUrl`, `userEmail`, `currentYear`, and `companyName` before they appear in attributes or body text.
> 
> A regression test feeds a malicious `userEmail` containing a crafted `mj-include` payload and asserts the included file content never appears in the output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b79b0312322002fef5034fd205f8ebacfc8aa4f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1597, opened 2026-09-03
fix(invites): require tenant admin for invitations (AYC-889)

sirolad:
> [!NOTE]
> **Medium Risk**
> Authorization changes on invite creation affect who can add members and assign roles; behavior is intentionally stricter and may block previously allowed flows.
> 
> **Overview**
> **Single-invite creation is now restricted to tenant administrators**, closing a gap where non-admin members could invite users (including other admins).
> 
> The HTTP `POST /api/invites` handler gains `@Roles(UserRole.ADMIN)`, aligning with bulk invite. **`CreateInviteWithSeatReservationUseCase`** enforces the same policy in the application layer: it reads the caller’s `role` and `orgId` from `ContextService` and throws `UnauthorizedInviteAccessError` unless the user is an admin for the target org—**before** seat lock or invite persistence, so internal callers cannot bypass HTTP guards.
> 
> Coverage adds a controller metadata test, a unit test for rejected non-admins, an E2E scenario (403 for a regular member, 201 for admin), and a `requestInvite` helper that returns the raw HTTP response. Module docs note the admin-only rule.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 720bf5f72099ef25811ccc09e53232f550b77944. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1614, opened 2026-09-07
fix(email): block MJML template injection (AYC-890)

sirolad:
Re-lands **PR #1598** (AYC-890). GitHub reports that PR as merged, but its content never reached `main`, and therefore never reached any release.

Production currently runs revision `2e5d6a505` = tag **v2.36.0** (from `appsignal.config.revision` on traces through 2026-09-07T10:38Z). `render-mjml.ts` is absent from v2.36.0 and from every earlier tag checked back to v2.33.0, so the guard is absent from the code actually deployed — not merely absent from `main`.

## Why the original fix is missing

#1598's base was the stacked parent `09-03-fix_invites_require_tenant_admin_for_invitations_ayc-889_`, not `main`. The order of events:

| Time (2026-09-03) | Event |
| --- | --- |
| `13:05:10Z` | #1597 (the parent, base `main`) squash-merged → `b1327e67c`. That squash contains **zero** email-template files. |
| `13:25:52Z` | #1598 merged into the parent branch — 20 minutes after the parent had already been squashed away. |

Merge commit `e7dd0406` is reachable only from the now-stale `origin/09-03-…ayc-889_`; `git merge-base --is-ancestor e7dd0406 origin/main` is false. On `main` today there is no `render-mjml.ts`, no `ignoreIncludes`, and all five templates call `mjml2html(...)` directly.

## Confirmed by execution, not inferred

Before applying anything, the reproduction spec was run against unpatched `main`. `main` is a strict descendant of the deployed v2.36.0 for these six files — all are byte-identical between the two — so this reproduction speaks for the deployed code:

```
● email confirmation template › renders an MJML include payload in an email address as inert text
  expect(rendered.html).not.toContain(includedMarker)
  Tests: 1 failed, 1 total
```

The assertion that failed is the one checking a local file's contents are **absent** from the rendered email. So on current `main`, an `mj-include` smuggled through an RFC-style quoted-local-part email address breaks out of the surrounding `mj-text` and causes `mjml2html` to read a local file and embed it in outbound mail. That is arbitrary local-file disclosure to an email recipient.

## Change

Cherry-picked unchanged from `b79b03123`; it applied to current `main` without conflict, because `main`'s versions of all six touched files are still identical to the pre-fix state the original was written against.

- **`render-mjml.ts`** (new) — wraps `mjml2html(source, { ignoreIncludes: true })`. All five templates render through it, so `mj-include` can no longer resolve local paths.
- **`email-confirmation.template.ts`** — escapes `confirmationUrl`, `userEmail`, `currentYear` and `companyName`. This was the one template interpolating user-controlled values into MJML source unescaped; the other four already went through the `_layout` helpers.
- **`_layout.ts`** — `escapeAttr` renamed to `escapeAttribute` and exported for the above.

Defence in depth: `ignoreIncludes` blocks the file read, escaping blocks the element breakout. Either alone would stop this specific payload; both are kept, as in the original.

## Verification

- **Red before green** — the spec failed on unpatched `main` (output above), passes after.
- **Full backend suite**: 748 suites, 5032 tests passed.
- `tsc --noEmit -p tsconfig.json` clean; `eslint --max-warnings=0` clean on all 8 changed files; all pre-commit gates green.
- Base is `main` this time, verified after submission — that was the whole failure mode last time.

Refs AYC-890, supersedes #1598.

### DIFF
--- a/ayunis-core-backend/src/common/email-templates/infrastructure/mjml/templates/_layout.ts
+++ b/ayunis-core-backend/src/common/email-templates/infrastructure/mjml/templates/_layout.ts
@@ -119,7 +119,7 @@ function renderBannerSection(
   return bannerUrl
     ? `<mj-section background-color="${COLOR.card}" padding="16px 32px 0 32px">
          <mj-column>
-           <mj-image src="${bannerUrl}" alt="${escapeAttr(bannerAlt ?? '')}" width="496px" border-radius="10px" padding="0" />
+           <mj-image src="${bannerUrl}" alt="${escapeAttribute(bannerAlt ?? '')}" width="496px" border-radius="10px" padding="0" />
          </mj-column>
        </mj-section>`
     : '';
@@ -220,6 +220,6 @@ export function escapeText(input: string): string {
     .replace(/>/g, '&gt;');
 }
 
-function escapeAttr(input: string): string {
+export function escapeAttribute(input: string): string {
   return escapeText(input).replace(/"/g, '&quot;');
 }
--- a/ayunis-core-backend/src/common/email-templates/infrastructure/mjml/templates/budget-warning.template.ts
+++ b/ayunis-core-backend/src/common/email-templates/infrastructure/mjml/templates/budget-warning.template.ts
@@ -1,7 +1,6 @@
-import mjml2html from 'mjml';
 import type { MJMLParseResults } from 'mjml-core';
 
-import type { BudgetWarningTemplateContent } from '../../../domain/email-template.entity';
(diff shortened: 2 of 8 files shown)


sirolad:
* **#1614** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1620, opened 2026-09-07
fix(anonymization): translate presidio spans to utf-16 offsets (AYC-905)

sirolad:
Closes AYC-905. Found while diagnosing AYC-902; independent of it.

## Problem

The pipeline mixed two string index spaces. `ayunis-core-anonymize/app/presidio_service.py` returns Presidio's `result.start` / `result.end` verbatim — Python string indices, which count **code points**, so an emoji is 1. The backend fed them to `substring` / `slice`, which count **UTF-16 code units**, where an emoji is 2.

So every span after an astral character resolved one unit short per preceding emoji, in `toDetection`, `apply-replacements.ts` and `apply-mask-replacements.ts` alike.

## What that actually did — measured, not reasoned

Same input through `PresidioAnonymizationProvider` → `AnonymizeTextUseCase` → `deanonymizeText`, with the HTTP client stubbed. Text `Hi 😀 ich bin der Dani`, one `PERSON` detection at code points 17..21 (code units 18..22):

| | before | after |
| --- | --- | --- |
| anonymized | `Hi 😀 ich bin der{{pii:PERSON_NAME_1}}i` | `Hi 😀 ich bin der {{pii:PERSON_NAME_1}}` |
| mask value | `" Dan"` | `"Dani"` |
| whitelisting `Dani` | still masked | correctly exempt |

Two harms, and the first is the serious one:

1. **PII partially left in the clear.** The mask covered code units 17..21, so the trailing `i` of `Dani` survived into the output while the preceding space was consumed. A privacy control leaking characters of the very value it exists to hide, and it leaks one more character per preceding astral character.
2. **Whitelist entries stopped matching.** `whitelist-filter.ts` compares an entry's pattern against `detection.text`, which was the mis-sliced `" Dan"`. A configured exemption for `Dani` therefore did not match, and the value was masked anyway.

**Correcting the ticket:** it claimed the stored mask value "will not round-trip through de-anonymization". That is wrong — `deanonymizeText` reproduced the input exactly even before the fix, because the token and the stored value were both derived from the same wrong offsets, i.e. self-consistently corrupt. Verified, not assumed. The round-trip was never the defect; the leak and the whitelist bypass are.

## Change

Translation happens in the adapter, where engine-specific results already become domain `PiiDetection`s.

- Build the code-point → UTF-16 offset table once per `detect` call. Its length replaces `countCodePoints` for the input-size limit, so the single pass that already measured the text now also translates its spans — no extra traversal, one function instead of two.
- Clamp an out-of-range span to the end of the text rather than trusting an external service's indices.

**Why the backend and not the Python service:** it keeps one conversion point, and it cannot break under a rolling deploy. Changing the service's contract would mean a window where an old backend talks to a converting service (or the reverse), silently double-converting or not converting at all. Converting here needs no contract change at all.

Safe to confine to this module: the offsets never leave it — `grep` finds only `replacements.length`, read for a log count in `execute-run-and-set-title`. `dropOverlappingResults` still compares raw code-point values, which is fine because the translation is monotonic.

## Verification

- **Red before green** — the new spec asserted `text: 'Dani', start: 18, end: 22` and failed with `text: " Dan", start: 17, end: 21`. A second test covers a span *before* any astral character, so the ASCII path is pinned as identity.
- **Full backend suite**: 748 suites, 5035 tests passed. The existing ASCII offset assertions (`start: 12, end: 16` for `Ich bin der Dani`) still pass unchanged.
- `tsc --noEmit -p tsconfig.json` clean; `eslint --max-warnings=0` clean on both changed files; all pre-commit gates green.

No end-to-end test was added on top: the composed harms both reduce to "the offsets and `detection.text` agree with the original string", which the adapter tests assert directly, and `applyMaskReplacements` / `whitelist-filter` are pure functions already covered. The table above is evidence, not a coverage gap.

Note on deployment: production runs tag **v2.36.0** (revision `2e5d6a505`); merging this does not ship it.

### DIFF
--- a/ayunis-core-backend/src/common/anonymization/infrastructure/providers/presidio-anonymization.provider.spec.ts
+++ b/ayunis-core-backend/src/common/anonymization/infrastructure/providers/presidio-anonymization.provider.spec.ts
@@ -143,6 +143,29 @@ describe('PresidioAnonymizationProvider', () => {
     expect(mockAnalyzeTextAnalyzePost).not.toHaveBeenCalled();
   });
 
+  // Presidio returns Python string indices, which count code points — an emoji
+  // is 1 — while substring/slice count UTF-16 code units, where it is 2. Every
+  // span after an astral character used to resolve one unit short per emoji,
+  // masking the wrong text and storing a mask value that cannot round-trip
+  // through de-anonymization (AYC-905).
+  it('resolves spans past an emoji using code-point offsets', async () => {
+    const text = 'Hi \u{1F600} ich bin der Dani';
+    mockResults([{ entity_type: 'PERSON', start: 17, end: 21, score: 0.9 }]);
+
+    const detections = await provider.detect(text);
+
+    expect(detections[0]).toMatchObject({ text: 'Dani', start: 18, end: 22 });
+  });
+
+  it('leaves spans before any astral character untouched', async () => {
+    const text = 'Dani schrieb \u{1F600}';
+    mockResults([{ entity_type: 'PERSON', start: 0, end: 4, score: 0.9 }]);
+
+    const detections = await provider.detect(text);
+
+    expect(detections[0]).toMatchObject({ text: 'Dani', start: 0, end: 4 });
+  });
+
   it('counts Unicode code points like the anonymize service', async () => {
     const text = '😀'.repeat(30_000);
     mockResults([]);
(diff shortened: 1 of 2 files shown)


sirolad:
* **#1620** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1620?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1620?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1633, opened 2026-09-08
feat(sources): accept .odt and .odp uploads (AYC-159)

sirolad:
OpenDocument text and presentation files were rejected on upload, so public-sector users had to convert them by hand before using them as a source. Both formats ride the existing document path: Gotenberg's LibreOffice endpoint reads ODF natively, so no new conversion infrastructure is needed.

Closes AYC-159.

## What changed

- `.odt` and `.odp` are accepted at all four upload entry points — chat, skills, knowledge bases and projects. Three of them share the `isDocumentSourceFile` gate, so extending `isDocumentFile` carries them; the frontend accept lists are updated per surface.
- `isOfficeDocumentFile` is now the single gate for the LibreOffice-to-PDF route. Previously `isDocumentFile` (upload) and a literal `docx || pptx` check (retrieval) could drift, letting a format be accepted on upload and then rejected during retrieval.
- The two duplicated mime-to-`FileType` switches collapse into one shared map (`source-file-type.helpers.ts`), so a new format cannot be recorded by one entry point and rejected by another. This also brings both `getFileType` methods back under the cyclomatic-complexity gate.
- `FileType` gains `ODT`/`ODP` plus an enum migration following the `AddEmlToFileTypeEnum` rename/add-value/cast pattern.
- `.odp` yields one text source per file, like PPTX — nothing slide-aware.
- Knowledge-base copy claimed "Supported formats: PDF, DOCX" (already stale before this change); it now names the formats the picker actually accepts.

Same 25 MB limit as every other source file — the cap is uniform in `SOURCE_FILE_UPLOAD_OPTIONS`, so no per-type change was needed.

## Verification

Unit: 750 suites / 5073 tests pass; lint, prettier, typecheck, complexity gate and file-size gate pass on the changed files.

Behaviour, on a dedicated dev slot (slot 5, real Bedrock/Azure models and real Mistral OCR — no mocked inference), seeded with `seed:minimal:ts`:

| Entry point | Endpoint | Result |
| --- | --- | --- |
| Chat | `POST /threads/{id}/sources/file` | 201, `fileType` `odt`/`odp`, both `ready` |
| Skill | `POST /skills/{id}/sources/file` | assigned in `skill_sources`, both `ready` |
| Knowledge base | `POST /knowledge-bases/{id}/documents` | 202, both listed `ready` |
| Project | `POST /workspaces/{id}/context/documents` (`FEATURE_WORKSPACES_ENABLED=true`) | 201, both `ready` |

Extracted text is structurally intact — the ODT yields markdown headings, a bullet list and a markdown table; the ODP yields per-slide text as a single source. The model answers from both files (screenshot below).

Migration: generated against a clean slot-12 database (zero unrelated drift), applied to a database holding existing source rows, and `migration:generate` afterwards reports "No changes in database schema were found".

Existing types re-checked on the same slot: TXT, DOCX and PPTX uploads all processed to `ready`.

One pre-existing bug surfaced during QA. It is **not** caused by this change and reproduces for DOCX just as much as for ODF, so it is filed separately rather than fixed here: a knowledge-base document intermittently loses its `knowledgeBaseId`. `assignSourceToKnowledgeBase` runs after `startDocumentProcessing` has already enqueued the job, and the consumer's full-record save (`source.mapper.createTextSourceRecord` writes `knowledgeBaseId`) can write back the value it loaded before the assign. Observed on `race-docx-1.docx` while all four ODT uploads in the same batch survived.

One ODT upload also failed once with `Mistral OCR processing failed` on the converted PDF. That is the OCR stage PDF/DOCX/PPTX share, the same file then succeeded 5/5 on retry, and `classifyJobFailure` settles a deterministic provider rejection on the first attempt by design (AYC-655). The error body is redacted in logs, so the upstream status is unknown — recording it as an observation, not a finding.

## Screenshots / demo

| ODF attached in chat | Model answering from the ODT table and ODP slides |
| --- | --- |
|  |  |

`.odt` documents in a knowledge base:

demo

> Media hosted on the `pr-media/ayc-159` branch (not part of this PR's diff); safe to delete after merge.

### DIFF
--- a/ayunis-core-backend/src/common/util/file-type.spec.ts
+++ b/ayunis-core-backend/src/common/util/file-type.spec.ts
@@ -2,7 +2,9 @@ import {
   detectFileType,
   getCanonicalMimeType,
   isAudioFile,
+  isDocumentFile,
   isEmailFile,
+  isOfficeDocumentFile,
   MIME_TYPES,
 } from './file-type';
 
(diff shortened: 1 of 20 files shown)


sirolad:
* **#1633** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1633?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1645, opened 2026-09-09
chore(deps): upgrade the OpenAI SDK to v7 (AYC-000)

sirolad:
openai 4.104.0 was three majors behind and declared a zod 3-only
optional peer (^3.23.8); v7 declares ^3.25 || ^4.0.

Our call surface is narrow — client construction, chat completions
streaming, embeddings and images — so the migration is four changes:

- images.edit() gained a streaming overload, so its return type is now
  a union. Declaring the params as ImageEditParamsNonStreaming selects
  the ImagesResponse overload we already return.
- input_fidelity is a first-class ImageEditParams field in v7, so the
  cast that forced it through the v4 typings is gone, along with the
  comment explaining the workaround.
- APIError.generate() takes a Headers instance rather than a plain
  object, so the image-generation spec passes new Headers().
- ChatCompletionTool is now a union of the function and custom tool
  shapes. convertTool only ever builds a function tool, so it returns
  ChatCompletionFunctionTool.

This does not remove the second zod major from the workspace:
@mistralai/mistralai declares zod as a hard dependency, so the
@hookform/resolvers packageExtensions entry is still required.

---

> [!NOTE]
> **Medium Risk**
> Major SDK bump on shared chat, embedding, and image paths; changes are small but affect production LLM and Azure image generation behavior.
> 
> **Overview**
> **Upgrades the `openai` package from v4 to v7** in `core-backend` and `@ayunis/provider-openai`, with lockfile refresh (v7 now targets Node ≥22 and broadens optional peers such as zod 3/4).
> 
> Call sites are adjusted for v7 typings and APIs: Azure reference-image edits use `ImageEditParamsNonStreaming` so `images.edit` resolves to the non-streaming `ImagesResponse` overload, and `input_fidelity: 'high'` is passed without the old type cast. Image-generation tests construct `APIError` with `new Headers()` instead of `{}`. The OpenAI provider’s `convertTool` return type is narrowed to `ChatCompletionFunctionTool` because `ChatCompletionTool` is now a union.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4c2f1416c89c578d7ee8b232d0d8bcb638a40d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Live QA

Verified against a real Azure/OpenAI environment (isolated dev slot, since torn down) — every changed line was exercised, not just typechecked.

| Chat via `openai()` — GPT-4o | Chat via `azure()` — GPT-5.4 |
| --- | --- |
|  |  |

Both constructors stream a real reply through `chat.completions`, exercising `convertTool`'s new `ChatCompletionFunctionTool` return type. The Azure thread also auto-titled, which is a second model round-trip.

| `images.generate` — gpt-image-1 | `images.edit` + `input_fidelity: 'high'` |
| --- | --- |
|  |  |

The right-hand shot is the one that matters for this diff: the follow-up edit ran with `referenceImageCount: 1` on `gpt-image-1`, so it took the `INPUT_FIDELITY_MODELS` branch — the exact line where the `as ImageEditParams` cast was removed. Composition, wall and shadow are preserved across the recolour, which is what `input_fidelity: 'high'` is for.

Not shown: a GIF. These are provider round-trips with no motion worth animating; the static frames carry the evidence.

> Media hosted on the `pr-media/pr-1645` branch, outside this PR's diff. It carries no `.pr-media/scenes.ts`, so the scene-capture job reports pr-media disabled rather than failing.

### DIFF
--- a/ayunis-core-backend/package.json
+++ b/ayunis-core-backend/package.json
@@ -105,7 +105,7 @@
     "node-html-parser": "^7.1.0",
     "nodemailer": "^7.0.5",
     "ollama": "^0.6.3",
-    "openai": "^4.104.0",
+    "openai": "^7.12.1",
     "openid-client": "^6.8.7",
     "otplib": "^12.0.1",
     "p-limit": "^7.3.1",
--- a/ayunis-core-backend/src/domain/models/infrastructure/image-generation/azure.image-generation.spec.ts
+++ b/ayunis-core-backend/src/domain/models/infrastructure/image-generation/azure.image-generation.spec.ts
@@ -112,7 +112,7 @@ describe('AzureImageGenerationHandler', () => {
           },
         },
         'content policy violation',
-        {},
+        new Headers(),
       );
 
       mockImagesGenerate.mockRejectedValue(apiError);
@@ -130,7 +130,7 @@ describe('AzureImageGenerationHandler', () => {
         500,
         { error: { message: 'server error', code: 'server_error' } },
         'server error',
-        {},
+        new Headers(),
       );
 
       mockImagesGenerate.mockRejectedValue(apiError);
--- a/ayunis-core-backend/src/domain/models/infrastructure/image-generation/azure.image-generation.ts
+++ b/ayunis-core-backend/src/domain/models/infrastructure/image-generation/azure.image-generation.ts
@@ -1,7 +1,10 @@
 import { Injectable, Logger } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { APIError, AzureOpenAI, toFile } from 'openai';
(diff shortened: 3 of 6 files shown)


sirolad:
* **#1645** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1645?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1645?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1654, opened 2026-09-09
chore(deps): pin shared runtime versions with a pnpm catalog (AYC-000)

sirolad:
Every dependency shared between a workspace package and the app that
consumes it was declared in two package.json files. When those drift,
pnpm keeps whatever the lockfile already pinned rather than re-resolving
a range it still satisfies, so the package quietly runs a different copy
than its consumer. Anything depending on a shared module instance then
breaks in a way that reproduces nowhere obvious:

- sonner: <Toaster> came from @ayunis/ui's 2.0.7 while toast() ran from
  the frontend's 2.0.8, so every toast in the app stopped appearing
- react-hook-form: 7.81 against 7.87 made Control<> from one copy
  unassignable to the other, 195 type errors
- @mistralai/mistralai: 2.5.0 against 2.6.4 split MistralError's class
  identity, which would have silently defeated the transient-error
  instanceof check and its retry

All three were fixed one at a time. A catalog removes the second
declaration site instead, so the versions cannot drift again.

peerDependencies are deliberately left alone. A package's peer range is
a compatibility floor — 'I work with anything here' — which is the
mechanism that prevents a second copy in the first place. Only the
dependencies and devDependencies entries move, and the devDependency is
what pinned the duplicate in each case above.

Dev tooling stays per-package: eslint, prettier, typescript and friends
disagree across the workspace today, but each package runs its own and
shares no module instance, so aligning them buys nothing.

Resolutions are unchanged apart from eventsource-parser collapsing from
two versions to one.

---

> [!NOTE]
> **Low Risk**
> Dependency and lockfile-only changes with no application code; intended to reduce duplicate-package bugs rather than change runtime behavior.
> 
> **Overview**
> Introduces a **pnpm workspace catalog** in `pnpm-workspace.yaml` as the single place to define versions for dependencies shared between apps and workspace packages (React stack, Tailwind, provider SDKs, MCP client).
> 
> **Backend**, **frontend**, **`@ayunis/ui`**, **provider-*** packages, and **agent-extensions** now declare those packages as `"catalog:"` instead of repeating `^` ranges in each `package.json`. **`peerDependencies` are unchanged**—only direct/dev dependency pins move to the catalog so consumers and libraries resolve the same module instance.
> 
> The lockfile is regenerated accordingly; **`@modelcontextprotocol/client`** also picks up minor transitive bumps (`eventsource-parser`, `jose`, `zod`) and drops a duplicate `eventsource-parser` version.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1b89489f0061ec8f06e3f4b396a49fb9fedadf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Verification

**Resolutions are unchanged.** The point of this PR is that nothing moves — I diffed every resolved version key in the lockfile before and after:

```
only difference:  eventsource-parser  2 versions -> 1
```

That one is an improvement (a transitive duplicate collapsed), not drift. Every other resolution is byte-identical.

**One copy of each cataloged dependency:**

```
react  react-dom  react-hook-form  sonner  recharts
lucide-react  tailwindcss  @mistralai/mistralai
openai  ollama  @modelcontextprotocol/client        1 copy each
@tailwindcss/vite                                   1 version, 2 peer variants
```

`@tailwindcss/vite` resolves to a single 4.3.3 with two peer-suffix variants, because the frontend builds with `vite@6.4.2` and `@ayunis/ui` with `vite@7.3.5`. Same package version compiled against different peers — harmless for a build-time Vite plugin, and `vite` itself is dev tooling this PR deliberately leaves per-package.

**Checks**

| | |
| --- | --- |
| frontend tsc / lint / tests / build | 0 errors · 0 errors, 368 warnings (baseline) · 136 files, 688 tests · ✓ built |
| backend tsc / suite | 0 errors · 5103/5106 — the 3 are the DB-dependent `register-user.integration.spec.ts`, identical on main |
| `packages/ui` tsc | 0 errors |
| all provider packages | pass (openai 71, mistral 34, anthropic 61, gemini 48, ollama 19, agent-extensions 55) |
| `pnpm audit --audit-level critical` | PASS |
| `pnpm install --frozen-lockfile` | in sync |

**Live smoke test** on the two things the duplication actually broke — toasts and the provider path:

```
✓ tests/admin/instructions.spec.ts        toggles org internet access   (success toast renders)
✓ tests/auth/account-lockout-message.spec.ts                            (toast text asserted)
✓ tests/chat/send-message.spec.ts         sends a message and receives the routed mock reply
✓ tests/chat/send-message.spec.ts         continues a conversation in an existing thread
```

The first two are the specs that failed on #1563 before the sonner copies were merged; the chat ones prove provider routing still works end to end.

### Follow-ups this does not do

- `pnpm dedupe --check` as a CI guard would turn this class of bug into a build failure. It **exits 1 today** with 252 dedupable entries across 42 packages, so it needs a reviewed `pnpm dedupe` pass first.
- The 11 dev-tooling range disagreements (`eslint`, `prettier`, `typescript`, `typescript-eslint`, `globals`, `knip`, …) are untouched — harmless, no shared module instance.


sirolad:
* **#1654** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1654?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1654?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1663, opened 2026-09-11
fix(integrations): refetch the list after disabling an integration (AYC-941)

sirolad:
Fixes AYC-941.

## Problem

Deactivating an integration on **Admin Settings → Integrations** showed a success toast but left the toggle visually enabled until a page refresh.

## Root cause

`useToggleIntegration` invalidated a hand-written query key on the disable path:

```ts
queryKey: ['useMcpIntegrationsControllerList']   // never matches
```

The generated list key is `['/mcp-integrations']`, so the invalidation matched nothing and the list never refetched. The enable path already used `getMcpIntegrationsControllerListQueryKey()` — only disable was wrong.

## Fix

Use the generated key on the disable path too. This was the only hand-written controller query key in the frontend.

## Acceptance criteria

- **Toggle updates immediately after activation or deactivation** — the card renders `checked={integration.enabled}` straight from the list query, so the now-matching invalidation refetches and flips it.
- **Rendered state stays consistent after refresh** — the refetch reads the same persisted state a reload would.
- **A failed request preserves the previous state and reports the failure** — there is no optimistic update, so a failure leaves server state untouched and the existing `onError` handlers show the error toast.

## Tests

`useToggleIntegration.test.tsx` asserts both enable and disable invalidate the generated list key, plus the toggling-state cleanup. The disable case fails on `main` and passes here.

### DIFF
--- a/ayunis-core-frontend/src/pages/admin-settings/integrations-settings/api/useToggleIntegration.test.tsx
+++ b/ayunis-core-frontend/src/pages/admin-settings/integrations-settings/api/useToggleIntegration.test.tsx
@@ -1,0 +1,71 @@
+import { act, renderHook } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { useToggleIntegration } from './useToggleIntegration';
+import type { McpIntegration } from '@/pages/admin-settings/integrations-settings/model/types';
+
+const LIST_QUERY_KEY = ['/mcp-integrations'] as const;
+
+type MutationOptions = {
+  onSuccess: () => void;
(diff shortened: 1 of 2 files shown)


sirolad:
* **#1663** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1663?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1663?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pavellnk (approved):

PR #1630, opened 2026-09-08
fix(letterheads): position letter content per DIN 5008 (AYC-711)

sirolad:
Letters generated onto a stored letterhead started at the configured top
margin (20mm by default), landing the body inside the letterhead's printed
header, with the sender's contact details in the flat single column instead
of beside the address.

- PDF export renders a header-less table as the DIN 5008 head block:
  recipient in the Anschriftzone at 62.7mm, information block at 125mm,
  subject at 98.46mm, single-spaced so a long address stays in its zone
- The marker class is injected at render time, after sanitization: the
  Tiptap schema drops classes and data attributes, so a marker stored in
  the content would vanish on the first edit
- DOCX mirrors the rule so the head block is not a visible grid
- The document tools now describe the letter structure to the model
- Letterhead margins default to DIN 5008 Form B, with Form A/B presets in
  the margin editor for adjusting existing letterheads

Verified by measuring text positions in a rendered PDF for Form A, Form B,
a 6-line address, and a mixed letter with a real data table.

---

> [!NOTE]
> **Medium Risk**
> Changes PDF/DOCX export layout heuristics (header-less tables) and default letterhead margins, which can affect existing exports and AI-generated letters; mitigated by tests and letterhead-only PDF layout CSS.
> 
> **Overview**
> **Letterhead PDF/DOCX export** now treats a **header-less two-column table** as the DIN 5008 address and sender block: PDF export runs `markLayoutTables` after sanitization (so TipTap cannot strip the marker), applies `letter-layout` CSS for zones, spacing, and subject alignment, and only when a letterhead is used. DOCX uses the same “no `<th>` ⇒ layout table” rule via `docx-table-styles` so the head block is borderless while real data tables stay gridded.
> 
> **AI document tooling** extends `buildLetterheadSuffix` with explicit letter structure guidance (two-column block, bold subject, headed data tables, no duplicated letterhead artwork).
> 
> **Admin UI** defaults new letterheads to DIN 5008 Form B margins, adds Form A/B one-click presets on the first-page margin editor, and continuation pages default to DIN continuation margins. The artifact editor previews header-less tables borderless to match export.
> 
> E2E CI turns on `FEATURE_LETTERHEADS_ENABLED`.
> 
[...]

PR #1682, opened 2026-09-15
fix(letterheads): surface loading and timeout states in the margin preview (AYC-711)

sirolad:
## What

**The load-bearing change is a real bug fix**, not the UX hardening the title suggests. On `main`, a failed PDF load leaves the preview permanently blank:

1. In the create dialog the File is local and unvalidated, so picking a corrupt or password-protected PDF rejects client-side and shows the error state.
2. The error branch **unmounted the canvas**, so `canvasRef.current` became null.
3. Picking a good PDF re-ran `renderPreview`, which hit `if (!canvas || !pdfSource) return;` — a silent resolve — and then cleared the error.
4. The canvas remounted at its unsized 300px default and the effect never re-ran. Blank preview, no error, until a full page reload.

`recovers when a good PDF replaces one that failed to load` covers this. Run against `main`'s component it fails with `expected 300 to be 280` — the canvas never gets sized. The rest of the change is defensible UX hardening on top.

Keeping the canvas mounted (the fix for the above) surfaced two adjacent defects, both caught by Bugbot and both fixed here:

- A failed replacement left the **previous** letterhead painted under the translucent overlay, so the preview contradicted the selected file. `renderPreview` now clears the canvas when the document changes — and only then, so a margin edit repainting a cached page does not flash.
- Status was reset during render while an in-flight load could still settle from a microtask before the effect cleanup ran, letting a superseded load clobber the replacement's loading state. Rather than patch that, the status is now **derived** from a settled result tagged with the source it describes (`settled?.source === pdfSource ? settled.status : 'loading'`). A stale settle is filed under the old source and ignored by construction. That removed a state variable and the render-phase reset block along with the race.

## Test coverage

8 tests, each mutation-checked against the code path it claims to cover:

| Test | Fails without |
| --- | --- |
| `recovers when a good PDF replaces one that failed to load` | the mounted canvas (fails on `main`) |
| `clears the old artwork when the PDF is replaced` | the `clearRect` on document change |
| `returns to loading when a PDF is selected in the create flow` | the source-tagged status |
| `ignores a settle from a source that has been replaced` | *(passes with either guard — encodes the contract, does not isolate the microtask window)* |
| `falls back to the error state when pdf.js never settles` | the timeout (brackets 14s / 16s) |
| `shows the error state when the PDF cannot be loaded` | the catch path |
| `shows the loading state while pdf.js has not resolved` | the overlay |
| `does not flicker back to loading when only the margins change` | the source comparison |

The letterhead margin preview had no feedback while pdf.js was working, and only a *rejected* load ever reached the existing "Could not render PDF preview" error state. A **pending** promise — a worker that never starts, a stalled PDF fetch — left an unsized, blank canvas forever, indistinguishable from a broken preview.

- Add a loading overlay (spinner + copy, EN/DE) and a 15s timeout into the existing error state.
- Throw instead of silently resolving when the canvas or its 2D context is missing, so those paths surface as the error state rather than a permanent blank.
- Keep the canvas mounted in every state. The error branch used to unmount it, so a retry after an error had no ref to draw into and silently blanked forever.

This matters because the preview is the only way an admin can visually confirm their margins line up with their letterhead artwork — AYC-711's DIN 5008 preset guidance assumes that check.

## What this is NOT

This does **not** fix a broken production render, because there isn't one. The originating report suspected the Vite `?url` pdf.worker import and missing `type: 'module'` handling. Both are ruled out:

- The worker asset **is** emitted and correctly referenced: `dist/assets/pdf.worker-B1D2UnXD.mjs` → `/assets/pdf.worker-B1D2UnXD.mjs`.
- `type: 'module'` is already handled by the library — pdfjs-dist 5.7.284 constructs `new Worker(workerSrc, { type: "module" })` itself (`build/pdf.mjs:15641`).
- The backend serves the asset as `text/javascript; charset=utf-8`; helmet's CSP declares no `worker-src`, so it falls back to `default-src 'self'` and the same-origin module worker is allowed.

The blank canvas in the PR #1630 PR-media capture was the pre-render window this change now fills — which also explains why no error state appeared: nothing had failed.

**This is now confirmed, not inferred.** The author of that capture scene reported first-hand that its `waitFor` gated on the canvas being *visible*, which is satisfied the instant the element mounts at its unsized 300x150 default — long before pdf.js paints into it. The published PNG corroborates it: the preview box carried the default's landscape aspect, not the ~280x396 A4 portrait a real render produces. A canvas that was never *sized* by `renderPreview`, not one that failed to paint.

Note for future scenes: `waitForScene` only asserts `toBeVisible()` on a returned `Locator`, so "wait until painted" cannot be expressed by returning one — a scene has to do that waiting inside its own `waitFor` body. The overlay this PR adds gives a clean generic signal for it: wait for `getByRole('status')` to disappear.

Three independent environments agree the render is fine: deployed staging (`core-stage.ayunis.de`, build `5eed2295`), production (`core.ayunis.com`, real customer letterhead), and the local prod-serve run below.

## Classification

**Standard Path.** Frontend-only behavior change on one widget. No public contract, persistence, auth, tenant boundary, migration, or provider configuration is touched, so no High-Risk trigger applies. Not Fast Path because it changes runtime behavior and adds logic rather than only copy or styling.

Failure modes covered: pending promise (timeout → error), rejected load (error), cancelled render (ignored, not an error), missing canvas/2D context (error, previously a silent resolve), error → retry (canvas stays mounted), and margin keystrokes re-rendering (no loading flicker — the status is deliberately not reset to `loading` on re-render).

## Validation

- `npx vitest run src/widgets/margin-editor` — 3/3 pass. The timeout test brackets 14s (no error yet) and 16s (error), so it proves the timer drives the transition. The first version of these tests passed vacuously because jsdom has no 2D context and the component errored before reaching pdf.js; `getContext` is now stubbed.
- `npx vitest run` — 785/785 pass, 168 files.
- `eslint --max-warnings=0` and `tsc --noEmit` clean on the changed files; full pre-commit suite green.
- **Real runtime:** production `vite build` copied into `ayunis-core-backend/dist/frontend`, served by the backend on `:3060` (slot 6, seeded DB, real letterhead, real `/api/letterheads/<id>/first-page-pdf`). The preview renders — canvas 280×396, ~69k non-white pixels, with the blue margin overlay. On remount the overlay showed `PDF-Vorschau wird geladen…` and then cleared to the render.

No PR media: the visible change is a transient loading overlay, and the steady state is unchanged from what PR #1630 already captured.


sirolad:
* **#1682** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1682?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1682?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1683, opened 2026-09-16
feat(workspaces): show skill and knowledge base counts in the list (AYC-000)

chrismahlkeui:
feat(workspaces): show skill and knowledge base counts in the list (AYC-000)

The workspace list only reported chats, which said nothing about whether
a workspace was equipped with skills or knowledge before opening it.

- Aggregate both counts per workspace in one batched query pair next to
  the existing chat stats, and expose them on the list response.
- The list row now reads skills and knowledge bases; zero values stay
  visible as a prompt to add resources.
- The chat count and the description move off the row, and a count is
  omitted while its feature toggle is off.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(workspaces): file workspace chats under their workspace in the sidebar (AYC-000)

Chats belonging to a workspace also showed up in the general chat list, so
the same chat appeared twice and the list said nothing about where a chat
belongs.

- The threads list accepts `unfiled` to return only chats outside any
  workspace. Like `workspaceId`, the parameter 404s while the workspaces
  feature is off.
- The sidebar gains a workspace group between favourites and chats. A
  workspace expands to its chats, which are loaded only once it is
  expanded, and links to its page when it holds more than the list shows.
- The general chat list asks for unfiled chats only, so a chat now appears
  either under its workspace or in the list, never in both.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(sidebar): let every sidebar group collapse from its label (AYC-000)

The chats group was already collapsible, but nothing on screen said so:
its trigger showed the label and the search link, no chevron. Favorites
and workspaces could not be collapsed at all.

- SidebarCollapsibleGroup renders a group label as a collapsible trigger,
  with a chevron right next to the label and any action pushed to the
  far side. Favorites, workspaces and chats all go through it, so the
  three groups behave and look the same.
- The chevron points right while the group is collapsed and down while it
  is open, matching the workspace rows inside the sidebar.
- It also stays out of the way: it fades in on hover and on keyboard
  focus, and stays visible while the group is collapsed, where it is the
  only hint that the group holds anything.
- useChatsSidebarOpen becomes useSidebarGroupOpen, which takes the
  storage key from the caller. The chats group keeps its existing key, so
  a collapsed sidebar stays collapsed after the update.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

fix(workspaces): stop favoriting a workspace on creation (AYC-000)

Creating a workspace added it to the favorites right away, so the
favorites list filled itself with everything the user had ever created
and stopped saying anything about what they actually wanted pinned.
Pinning is now the user's decision alone, through the existing star.

Creation was the workspaces module's only reason to reach into favorites,
so the module import goes with it. The use-case spec no longer provides
AddFavoriteUseCase: putting the dependency back would fail every test in
the file on DI resolution.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(sidebar): expand a favorited workspace into its chats (AYC-000)

A workspace pinned to the favorites was a dead end: the same workspace
one group below could be expanded into its chats, the pinned one could
not, so pinning cost the user a feature.

The toggle and the chat list move into WorkspaceChatsToggle and
WorkspaceChatsSubMenu, shared by both rows. On a favorited workspace the
row menu shifts left to make room for the chevron; favorited chats keep
the row they had.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(sidebar): give workspace rows a row menu and a quieter chevron (AYC-000)

Every workspace row showed a permanent chevron, which added one arrow per
workspace to a sidebar that already has three group chevrons. The rows
also had no actions at all, while the same workspace pinned to the
favorites offered a full menu.

- The chevron follows the sidebar's usual row affordance: hidden until
  the row is hovered or focused, and kept visible while the chats below
  it are open.
- Workspace rows gain the row menu next to it, with edit, pin or unpin,
  and delete. Pinning matters more now that creating a workspace no
  longer pins it by itself.
- The settings and delete dialogs live in the group, mirroring how the
  favorites group hosts the same two dialogs.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

fix(sidebar): mark only one row as the current workspace (AYC-000)

Opening a workspace lit up three rows at once: the "Arbeitsbereiche"
section, the workspace under favorites, and the same workspace in the
workspaces group. The highlight stopped answering "where am I".

- The section item matches its own route exactly, the way "Neuer Chat"
  already does. Skills and knowledge keep prefix matching, since their
  detail pages have no row of their own in the sidebar.
- A pinned workspace no longer repeats in the workspaces group. The chats
  group has always dropped pinned chats the same way; the group hides
  itself once every workspace is pinned.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

fix(sidebar): keep the row menu anchored to the row edge (AYC-000)

Making room for the chevron moved the row menu 24px inwards, and the
dropdown follows its trigger: on a workspace row the menu no longer
lined up under the row the way it does everywhere else.

The row menu goes back to the outer edge, the same position it holds on
chat rows, and the chevron takes the inner slot instead.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(sidebar): put the workspace chevron next to the name (AYC-000)

The chevron sat in the row's action slot on the right, far away from the
name it belongs to, while the group labels carry theirs right behind the
text. Workspace rows now read the same way.

Keeping the whole row clickable while a second control sits inside it
uses the pattern the workspace overview already relies on: the link
stretches across the row through after:inset-0 and the chevron sits on
its own stacking context. The row menu keeps the action slot to itself.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

fix(sidebar): stop row menus from hugging the window edge (AYC-000)

The row menus are wider than the sidebar and align to their trigger near
its right edge, so they reach past the sidebar. Radix then keeps them
inside the viewport with no gap at all, and the menu ends up flush
against the window.

All three row menus now ask for 8px of collision padding.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

fix(workspaces): call a knowledge base a Wissenssammlung (AYC-000)

The German workspace screen was the last place calling them
"Wissensdatenbanken" while the chat, the sidebar and the knowledge pages
all say "Wissenssammlung". Ten strings on that screen now match, empty
state included.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---

> [!NOTE]
> **Medium Risk**
> Changes thread listing semantics and sidebar navigation across workspaces; API adds gated query params and new list fields, but scope is mostly UX and read paths rather than auth or data mutation.
> 
> **Overview**
> This PR extends **workspaces and sidebar navigation** so chats are filed under workspaces in the UI, workspace lists show resource counts, and pinning behavior is more intentional.
> 
> **Threads API:** `GET /threads` gains an **`unfiled`** query flag (feature-gated like `workspaceId`) that limits results to chats with no workspace. The sidebar’s general chat list uses this so workspace chats only appear under their workspace.
> 
> **Workspace list API & UI:** List responses add **`skillCount`** and **`knowledgeBaseCount`** (batched next to existing chat stats). Rows show those counts (with zeros kept visible); chat count and description are removed from the row. **Creating a workspace no longer auto-favorites it**—the favorites module dependency is dropped from workspaces.
> 
> **Sidebar:** A **workspaces group** sits between favorites and chats, with expandable rows that lazy-load workspace chats via `useWorkspaceThreads`. Favorited workspaces can expand the same way; pinned workspaces are omitted from the unpinned list (mirroring pinned chats). **`SidebarCollapsibleGroup`** + **`useSidebarGroupOpen`** unify collapse behavior for favorites, workspaces, and chats. Workspace rows get edit/pin/delete menus, quieter chevrons, exact route highlighting for “New chat” / Workspaces nav items, and **8px dropdown collision padding** on row menus.
> 
> **Chat lists:** **`sortFavoritesFirst`** sorts pinned threads first on the chats overview and workspace chats tab.
> 
> **Copy:** German workspace strings align on **Wissenssammlung**; new sidebar i18n for workspace chat expansion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c9b2c73174b527faa6603680abb7894b4dfed78. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chrismahlkeui:
* **#1683** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1686, opened 2026-09-16
refactor(skills): take pinning off the surface (AYC-000)

chrismahlkeui:
refactor(skills): take pinning off the surface (AYC-000)

Pinning only ever put a shortcut button on the new-chat screen, and that
row grew without a limit: twenty pinned skills meant twenty buttons under
the composer. Almost nobody used it, and the skills it advertises already
activate themselves through their triggers.

- The pinned-skills row leaves the new-chat screen, together with the skill
  selection it fed.
- The pin control leaves the skills list, the skill detail page and both
  workspace equivalents, so no screen offers an action the product no
  longer acts on.
- Two onboarding steps taught exactly this workflow and spotlighted the
  elements that are gone; they go with it, along with their tour targets
  and copy.

The endpoint, the response field and the database column stay untouched:
this is the surface only, and a later picker may well order by the flag.
`useSetSkillPin` and `setSkillPinned` therefore remain as the client for
that endpoint, covered by the existing contract tests.

Note for review: with the row gone there is currently no way to attach a
skill by hand — that is the next step, not an oversight. This branch
should not land on its own.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): offer three ways to add a skill (AYC-000)

Adding a skill meant one thing: filling in a form. Writing a good trigger
is the hard part of that form, and the marketplace was mentioned in a
hint below it rather than being reachable.

The button opens a menu instead:

- "Mit Ayunis Core erstellen" opens a new chat with a starting prompt
  that has Ayunis Core interview the user — what the skill is for, when
  it should kick in, what it should avoid — and create it at the end
  through the existing create_skill tool, which already renders a
  confirm-before-save widget.
- "Selbst erstellen" opens the editor dialog that used to sit behind the
  button.
- "Aus dem Marketplace hinzufügen" opens the marketplace in a new tab,
  and only appears where a marketplace is configured.

The empty state offers the same three routes, so the first skill is not
funnelled into the form either. Opening the editor from a menu item uses
the dropdown-to-dialog handoff the sidebar already relies on, so focus
does not jump back to the closed menu.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): let Ayunis Core sharpen a trigger or its instructions (AYC-000)

Triggers decide whether a skill ever activates, and they are the field
people fill with "whenever relevant". The form now offers to rewrite
either text, and each rewrite sees both fields: a thin trigger is derived
from what the instructions describe, and vague instructions are sharpened
against the situation the trigger names.

- POST /skills/improve-text runs a single inference through the caller's
  own default model, the same path the chat-title generator uses. It
  returns text only; nothing is stored and nothing is applied server-side.
- The button sits next to each field label, stays disabled while its own
  field is empty — there is nothing to improve about an empty box — and
  offers to put the previous wording back once a suggestion lands.
- While it runs the field carries a soft pulse and the icon twinkles, both
  silenced under prefers-reduced-motion.

Verified against the running stack: "Immer wenn relevant" came back as
"Wenn es um einen Bauantrag geht und geprüft werden soll, ob Fristen nach
der Landesbauordnung eingehalten wurden …", derived from the instructions
alone.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): explain the trigger and the instructions where they are written (AYC-000)

Both fields ask for something the form never explains, and the trigger is
the one that decides whether a skill ever fires. Each label now carries a
question mark that opens the long version, with an example from the kind
of work our users actually do — a concrete situation for the trigger, and
approach, tone and what to leave out for the instructions.

The improve action moves to the brand colour, which is a colour token on
the existing ghost button rather than a new button variant, and its
tooltip now just says what it does instead of describing the mechanism.

InfoHint moves from the roles settings into shared/ui: it has no domain
knowledge, and this is its second consumer. Its trigger label travels
with it into the common namespace.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): decide what happens to a suggestion (AYC-000)

A suggestion used to arrive with a single way out. Once it lands, the row
now offers all three: keep it, ask for another one, or put the original
wording back. Retrying starts from the original text rather than from the
suggestion, so the second attempt is an alternative instead of a rewrite
of a rewrite.

While a field is being worked on it now carries the ambient mist of the
new-chat backdrop — same brand colour, blurred and slowly breathing,
scaled down to the size of a control — instead of the harder ring pulse,
and both fields behave identically in the dialog and on the detail page.

The trigger explanation loses its closing warning; the example already
shows what a usable trigger looks like.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): show that Ayunis Core is working on a field (AYC-000)

While a trigger or an instruction is being rewritten, the field wears the
brand light the chat input shows during a run, and the button's label
shimmers. Both fields behave the same, in the create dialog and on the
detail page.

The effect is its own piece in shared/ui rather than a second copy of the
chat's: the chat rotates a circular element behind its card, which only
reads on a box roughly as tall as it is wide. Here the ring paints itself
— a conic gradient with an animated angle — so the light travels the
whole perimeter at any shape. The control drops its own border while the
ring is lit, the way the chat input does, otherwise the grey frame
swallows it. One lap takes 3s rather than the chat's 5.2s, because these
calls are seconds rather than a whole answer.

Two things that block the effect from ever being seen, both fixed here:
each improve button reported its pending state into one shared flag and
cleared it on every render, so the instructions button wiped the
trigger's glow in the same tick; and the trigger was a one-line input in
the dialog, where the lit segment crosses the long edges in an instant.
It is a textarea now, the shape it already had on the detail page.

Rewritten instructions also came back too short for what a skill carries
in the marketplace. The prompt now asks for roughly 150 to 400 words
covering role, approach, tone, layout, what to ask back and what to leave
out — and draws the line that matters: when the author was brief, expand
on how to work, never on what is true.

Tests drive both surfaces through the real mutation and check the field
still carries the glow once the effects settle; reverting the shared-flag
fix fails the trigger case and passes the instructions one, which is
exactly how it showed up.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

chore(skills): drop what the detours left behind (AYC-000)

- The settled-flash animation was written for a confirmation step that
  never shipped and had no caller.
- ProcessingGlow took a className nobody passed, and the improve button a
  disabled flag nobody set; both fell out of earlier drafts.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): show what a new chat can draw on (AYC-000)

The start screen never said which skills or knowledge were in play. The
old pinned row only showed the handful someone had bookmarked, and
nothing said where a skill came from or how many there were at all.

Two counters now sit on the workspace row, next to the picker that
changes them: skills and knowledge. Each opens into a scrollable list of
names grouped by origin — the picked workspace first, then your own, then
what others shared with you. Picking a workspace changes both numbers in
place, which is the cheapest way to show that a workspace brings its own
along.

Skills count only while they are active, so the number matches what the
chat will actually do rather than what exists somewhere. Each counter
appears only where its feature is switched on.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(new-chat): lift the privacy note above the greeting (AYC-000)

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): let the count open into a picker (AYC-000)

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

feat(skills): offer the way to manage them from the chat (AYC-000)

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---

> [!NOTE]
> **Medium Risk**
> Adds an authenticated LLM-backed endpoint and changes core new-chat and skill-editing flows; pinning removal is UI-only but affects how users attach skills until the new picker ships.
> 
> **Overview**
> This PR reshapes how users discover skills on new chat and how they author them, while **removing pin controls from the product surface** (pinned row, cards, detail, workspace). The pin API and `isPinned` field stay on the backend for a possible later use.
> 
> **New chat:** `PinnedSkills` is replaced by **`AvailableContextMenus`** — skill and knowledge counters beside the workspace picker that open scrollable lists (workspace + personal, deduped, active skills only, shared badges). Users can **toggle a skill** for the chat from that menu; knowledge stays read-only with hints. The privacy note moves **above the greeting**.
> 
> **Skill creation:** **CreateSkillMenu** offers guided chat (prefilled prompt), self-serve editor dialog, and optional marketplace link. The create dialog can be opened in a controlled way from the menu.
> 
> **AI-assisted authoring:** **`POST /skills/improve-text`** (`MANAGE_SKILLS`) runs one inference on the caller’s default model and returns rewritten **trigger** or **instructions** (both fields sent in the prompt). The UI adds **Improve** on create/edit forms with keep / retry / undo, field **InfoHint** copy, and a shared **`ProcessingGlow`** while a rewrite runs.
> 
> **Cleanup:** Pin-related onboarding steps and tour targets are removed; **`InfoHint`** moves to `@/shared/ui/info-hint` with common i18n for the trigger label.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 153b7376f8f7bf24830051a52cf622f6e0973b6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chrismahlkeui:
* **#1687** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1686** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1688, opened 2026-09-16
ci(build-images): stop sourcemap upload failures from blocking staging deploys (AYC-000)

sirolad:
## Problem

The `fix(sources): start PDF timeout after queue pickup (AYC-975)` merge (`472fee1`) never reached staging. All four images built and merged to GHCR successfully, but the `sourcemaps` job died on a transient AppSignal `curl: (22) ... error: 502`:

```
Uploading ./assets/knowledge-base-list-item-CKGp-uJ0.js.map ... (env: internal)
curl: (22) The requested URL returned error: 502
##[error]Process completed with exit code 22.
```

One failed job made the whole *Build & Publish Images* run conclude `failure`, so `deploy-staging`'s guard

```yaml
if: github.event.workflow_run.conclusion == 'success' && ...
```

skipped the deploy ([run 35103492241](https://github.com/ayunis-core/ayunis-core/actions/runs/35103492241) — `completed/skipped`). `notify-infra` already carries `continue-on-error` for exactly this reason; `sourcemaps` never got the same treatment.

## Fix

The job had two failure classes collapsed into one step. They are split:

| | Step | Gating? | Why |
| --- | --- | --- | --- |
| Our regression | **Verify the build produced sourcemaps** | **yes** | The `no sourcemaps found in dist` assertion is the *only* check in the repo guarding `sourcemap: 'hidden'` in `vite.config.js`. Nothing else in CI greps for `*.js.map`. Deploying a build that silently stopped emitting maps would leave every later backtrace minified, so this still fails the run. |
| Their uptime | **Upload sourcemaps to AppSignal** | no | A 502 costs readable backtraces for one revision. It must not skip a deploy. Failures are counted and reported as a `::warning::` annotation, so a green run still says what was lost. |

Upload requests are bounded and retried: `--connect-timeout 10 --max-time 30 --retry 5 --retry-delay 2 --retry-max-time 120`, with `timeout-minutes: 20` on the step. Once one upload exhausts its retries the loop `break 2`s out rather than multiplying an AppSignal outage across every remaining map and environment.

`--retry-all-errors` is deliberately *not* used: plain `--retry` already treats HTTP 502 — the error that caused this incident — as transient, and leaving it off avoids burning five retries on a genuine 401 or 413. Verified below.

A blanket job-level `continue-on-error` was the first attempt and was rejected: it would have swallowed the sourcemap-regression signal along with the outage.

## Classification

**High-Risk Path** (trigger: infrastructure/CI change).

### Failure-mode matrix

| Transition | Before | After |
| --- | --- | --- |
| All uploads succeed | run success → deploy runs | unchanged |
| Transient 502 on one upload | job fails → run failure → **deploy skipped** | curl retries 5×; recovers silently |
| 502 outlasting the retries | job fails → run failure → **deploy skipped** | `break 2` stops the loop, one `::warning::`, run success → deploy runs |
| AppSignal fully down | run failure → **deploy skipped** | first upload exhausts retries and breaks out (~10s, not ~10min), run success → deploy runs |
| AppSignal hangs without responding | job runs to the 6h default | `--max-time 30` per request, `timeout-minutes: 20` on the step |
| **Build emits no `.js.map`** | `::error::` + `exit 1` → run failure | **unchanged — still fails the run and blocks the deploy** |
| `APPSIGNAL_PUSH_API_KEY` unset | gate skips all steps, run success | unchanged |
| `pnpm install` / `pnpm run build` fails | run failure | unchanged (still gating; also covered by `frontend-build.yml` and the image build) |
| Image build job fails | run failure → deploy skipped | unchanged — image jobs remain gating |
| Tag / release-please build | deploy job's `head_branch == 'main'` check applies | unchanged |

### Ownership trace

- **Trigger/transport** — `deploy-staging.yml` `workflow_run` guard: untouched, still keys off run conclusion.
- **CI** — `build-images.yml` `sourcemaps`: upload tolerant, map-existence assertion still gating. The four `build` and `merge` jobs unchanged, so a broken image still blocks the deploy.
- **Runtime/persistence** — none. No application code, schema, or host state.
- **Telemetry** — worst case is minified frontend backtraces in AppSignal for one revision; errors are still reported and `ignoredErrorPatterns` behavior is unaffected.

### Validation

**1. Plain `--retry` does retry a 502.** The step's exact curl flags were run against a local server returning a permanent 502, counting requests server-side:

```
curl: (22) The requested URL returned error: 502   (x6)
curl EXIT=22    HITS=6    10.0s total
```

6 requests = 1 initial + 5 retries. The incident's failure mode is still covered without `--retry-all-errors`.

**2. Step bodies behave as intended.** Both `run` blocks were extracted from the parsed YAML (`js-yaml` → `jobs.sourcemaps.steps[].run`) and executed against a stubbed `curl`, with three maps sorting as `aaa` (ok), `bad` (fails), `zzz` (would succeed):

```
--- upload: aaa ok, bad fails, zzz should be SKIPPED by break 2 ---
Uploading ./assets/aaa.js.map as assets/aaa.js (env: staging)
Uploading ./assets/aaa.js.map as assets/aaa.js (env: production)
Uploading ./assets/bad.js.map as assets/bad.js (env: staging)
curl: (22) 502
upload failed: assets/bad.js (staging)
Uploaded 2 sourcemap(s) for revision abc123
::warning::sourcemap uploads stopped after 1 failure for abc123 - remaining uploads were skipped and backtraces may stay minified in AppSignal
EXIT=0
--- verify: no maps present (must exit 1) ---
::error::no sourcemaps found in dist - is build.sourcemap disabled?
EXIT=1
```

Confirms `set -euo pipefail` does not abort on the `if curl` condition, `break 2` short-circuits both loops, the step still exits 0, the summary annotation fires, and the map-existence gate still fails closed.

**3. The job was executed on a real runner.** `build-images.yml` has no `pull_request` trigger, so no PR check exercises this job — it was dispatched on the branch instead ([run 35148841078](https://github.com/ayunis-core/ayunis-core/actions/runs/35148841078), `workflow_dispatch`, `573c05d`):

```
success  sourcemaps
  success  Verify the build produced sourcemaps
  success  Upload sourcemaps to AppSignal
Uploaded 777 sourcemap(s) for revision 573c05ddcaa3138d27caafeb40630ac7013f9cc8
```

No warning annotation, no failed uploads. Safe to dispatch from a branch because `type=raw,value=latest` is gated on `github.ref == 'refs/heads/main'`, so only `sha-573c05d` was published — nothing staging or production pulls — and `deploy-staging` did not fire (`workflow_run.head_branch != 'main'`).

That run concluded `failure` for an unrelated reason: both `build (ayunis-core-python-sandbox, *)` legs hit `Failed to FinalizeArtifact: (403) Forbidden: Error from intermediary` in `actions/upload-artifact`, after their content had uploaded. This PR's only hunk is `@@ -346,31 +346,60 @@`, entirely inside the `sourcemaps` job (line 303+); the `build` matrix is untouched.

### Known tradeoff

`break 2` treats the first exhausted failure as an outage signal. A single pathological map (one oversized file returning 413) therefore also skips every map after it. Accepted: sourcemaps are best-effort telemetry, partial coverage is already the degraded state, and the alternative is per-error-class branching that is not worth the complexity here.


sirolad:
* **#1688** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1688?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1688?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1692, opened 2026-09-17
fix(auth): restore browser credential autofill on the auth forms (AYC-982)

sirolad:
## Problem

A customer reported that since email-first SSO login shipped, their browser no longer offers their saved email on the login page, so they can't click to sign in quickly. The initial read was that this is a browser credential-storage issue rather than an app issue. It isn't — it's our markup.

## Root cause

Before email-first discovery (`50d7fc9b7^`), the login `<form>` contained the email input immediately followed by the password input. Every password manager recognises that shape by heuristic and offers saved credentials on the email field, with no attributes required.

Email-first split the flow: step 1 renders a form containing **only** the email input — the password field isn't in the DOM at all. The heuristic no longer fires, and a grep showed **no `autocomplete` attribute anywhere in the auth forms** to fall back on. Chrome requires `autocomplete="username"` to offer password-manager credentials on a single-field username step, so the user gets at best plain address-profile email autofill. Exactly the reported symptom.

A second, related defect on the same path: step 2 rendered the email input as `disabled`. Disabled fields are dropped from form parsing, so a manager can't reliably pair the username with the password it's offered to save. `readOnly` keeps the field non-editable while leaving it parseable.

## Changes

| Form | Added |
| --- | --- |
| Login | `username` on email, `current-password` on password; step 2 locks the email with `readOnly` instead of `disabled` |
| Register | `username` on email, `name` on the user name, `new-password` on password |
| Invite accept | `username` on email (and `readOnly` instead of `disabled`), `name`, `new-password` |
| Reset password / activation | `new-password` on both password fields |
| Account password settings | `current-password`, then `new-password` on both new-password fields |

Two files being touched also had `../` imports, converted to `@/` path aliases per the repo import rule.

Deliberately **not** included:
- `autocomplete="organization"` on register's workspace-name field. It's the org being *created*, not the user's employer, so filling it from the address profile is more likely wrong than helpful.
- A hidden username field on the reset-password form. That flow is token-only and never has the email in scope.

## Classification — Fast Path

Trigger: isolated, easily reversed, no change to any public contract, persistent data, security boundary, or cross-module interaction. Each form's behaviour is demonstrable at one layer.

Not High-Risk despite touching auth pages: nothing in the authentication, authorization, or session path changes. These are presentation attributes plus one input-locking mechanism swap (`disabled` → `readOnly`) that alters no submitted value — the email is a controlled react-hook-form field either way, and `readOnly` blocks user editing identically. No credential handling, validation, or tenant scope is affected.

### Failure-mode matrix

| Transition | Risk | Check |
| --- | --- | --- |
| Step 1 renders | Manager still not offered the credential | New test asserts `autocomplete="username"` |
| Step 1 → step 2 | Password field not recognised | New test asserts `current-password` |
| Step 2 email locked | `readOnly` lets the user edit the email, diverging from the discovered SSO result | New test asserts `readonly` present and `disabled` absent; the value stays a controlled RHF field |
| Discovery in flight | Email editable mid-request | `disabled={isDiscovering}` retained; pre-existing test still passes |
| Change email → re-discover | Stale lock or stale password | Pre-existing "clears the password when the email is changed" test passes |
| Invite email switched to `readOnly` | Prefilled email becomes user-editable | Field stays a controlled RHF value; `readOnly` blocks input |
| Register / reset submit | Wrong credential saved | Attributes only; submission payload untouched, existing submit tests pass |

## Validation

Environment: local, `ayunis-core-frontend`, `main` @ `0fb13d12a`.

```
bun x vitest run src/pages/auth src/pages/settings/account-settings   # 8 files, 29 tests → 30 with the new one, all pass
bun x eslint --max-warnings=0 <6 changed files>                        # clean
bun x tsc --noEmit -p tsconfig.json                                    # clean
bunx prettier --check <6 changed files>                                # clean
```

Full pre-commit suite passed on the staged set: Prettier, ESLint, TypeScript, dependency-cruiser (no circular deps), duplication, file-size.

**The new test was confirmed to be real proof** — reverting the fix (stripping `autoComplete` and restoring `disabled`) makes it fail with `expected null to be 'username'`, and restoring the fix makes it pass again.

No PR media: the only visual delta is the step-2 email keeping its muted background via `className`, which no reviewer needs a screenshot to evaluate.

## Follow-up worth noting

The real end-to-end proof — a browser actually surfacing a stored credential — can't be asserted in jsdom or reliably in Playwright, since it depends on the browser's own credential store. Worth a manual pass in Chrome with a saved credential before release; the attribute contract is what's locked by tests here.

---

> [!NOTE]
> **Low Risk**
> Presentation-only HTML attributes and disabled→readOnly on email fields; no changes to auth logic, validation, or submitted payloads.
> 
> **Overview**
> Restores browser and password-manager autofill on auth flows by adding standard **`autocomplete`** hints (`username`, `name`, `current-password`, `new-password`) across login, register, invite accept, reset/activation, and account password settings.
> 
> On **email-first login**, step 2 no longer **`disabled`** the email field (which broke credential pairing); it uses **`readOnly`** with muted styling while discovery still **`disabled`** the field when pending. Invite accept applies the same **`readOnly`** swap for the prefilled email.
> 
> A new **login test** locks the username/current-password attributes and that step-2 email stays **`readonly`** but not **`disabled`**. Two pages switch relative imports to **`@/`** aliases only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c86a8e4d25c44b23e6385088415009949184e115. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sirolad:
* **#1692** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1707, opened 2026-09-21
fix(subscriptions): manage the newest subscription, not the active one (AYC-995)

sirolad:
> **Re-targeted from #1698.** That PR was stacked on #1695 and merged **into `cursor/ayc-977-subscription-history-0a38` at 11:58**, 38 minutes *after* #1695 had already merged into `main` at 11:20. GitHub marks it merged — it merged into a branch that was by then a dead end, so none of this code reached `main`.
>
> This branch is the same commit cherry-picked onto current `main`. The payload is byte-identical to what #1698 merged (`git diff 628a493c8 90da43c21` == `git diff main HEAD`): 13 files, +551/−73. No review comments from #1698 are outstanding — both findings there were fixed and resolved before it merged.
>
> Re-verified against current `main` (which has since gained AYC-992's ContextService refactor): `tsc --noEmit` clean, 152 subscription tests pass.

---

Stacked on #1695 (AYC-977). Closes AYC-995.

## Summary

Three super-admin-only subscription mutations resolved the organization's subscription through `GetActiveSubscriptionUseCase`, which filters by `isActive`. A subscription whose `startsAt` is in the future is not active, so the lookup found nothing and every one of them returned `404 SUBSCRIPTION_NOT_FOUND` — even though the super-admin view renders their controls on exactly that record.

- `DELETE /super-admin/subscriptions/:orgId` (cancel)
- `PUT /super-admin/subscriptions/:orgId/billing-info`
- `PUT /super-admin/subscriptions/:orgId/monthly-credits`

They now resolve through a new `findManageableSubscription` utility, which returns the newest record — the same one `GET /super-admin/subscriptions/:orgId`, `ChangeSubscriptionUseCase` and `UncancelSubscriptionUseCase` already use. The inconsistency between the two groups was the bug.

Found during pre-merge QA of #1695. Not a regression from it — the behaviour predates it — but that PR's history view is what surfaced it.

## Seats deliberately excluded

`PUT /super-admin/subscriptions/:orgId/seats` has the same 404 but is **not** changed here. `UpdateSeatsUseCase` is also driven by the org-admin invite auto-bump, which derives the new count as `noOfSeats + 1` from the *serving* subscription. Retargeting it at the newest record would raise a scheduled contract's seats while leaving the serving licence full — letting invites past the licensed count and overwriting a future contract's seat count (a 50-seat scheduled contract would become `servingSeats + 1`). Caught by the Cursor Security Agent on the first revision of this PR; the use case is reverted, pinned with a regression test, and documented at the call site. Seats on a scheduled subscription therefore still 404 — tracked separately.

## Deliberately unchanged

These paths still return `400 MULTIPLE_ACTIVE_SUBSCRIPTIONS` when an organization has more than one currently serving subscription. That leaves such an organization unmanageable through the UI, which is the AYC-938 data state #1695 made visible — but changing it is a product decision, not a refactor, so `findManageableSubscription` preserves the refusal. Worth a follow-up ticket.

## Workflow

**High-Risk Path** — billing, and a shared lookup behind a super-admin API contract.

### Failure-mode matrix

| Transition | Check |
| --- | --- |
| Scheduled subscription (`startsAt` in future) | All three mutations succeed — was 404 |
| Seat update with a newer scheduled subscription | Still writes the serving record; scheduled seats untouched |
| Several subscriptions, newest scheduled | Acts on the newest, not an older serving one |
| More than one currently serving | Still `MultipleActiveSubscriptionsError`, nothing written |
| No subscription at all | Still `SubscriptionNotFoundError` |
| Already cancelled | Still `SubscriptionAlreadyCancelledError` |
| Admin of another organization | Still `UnauthorizedSubscriptionAccessError`, no repository call |
| Wrong subscription type | Seats still reject usage-based; credits still reject seat-based |

### Ownership trace

Super-admin controller → the four use cases → `findManageableSubscription` → `SubscriptionRepository.findByOrgId`. Each use case already called `validateSubscriptionAccess` itself before the lookup, so dropping the `GetActiveSubscriptionUseCase` dependency does not remove an authorization check — asserted by a test. The three changed use cases are reachable only through `super-admin-subscriptions.controller.ts`; no org-admin path changes. `UpdateSeatsUseCase`, which invite creation reaches, keeps active-only semantics and its only diff against the parent branch is a comment.

## Validation

Backend unit tests: 147 passed in the subscriptions module, including a new `CancelSubscriptionUseCase` spec covering every row of the matrix above.

Focused E2E (`scheduled-subscription-management.spec.ts`) plus the whole super-admin suite: 11 passed against a live stack.

Live runtime verification on a scheduled subscription, the same probes that produced the original 404s:

```
create scheduled sub      -> HTTP 201   (status: SCHEDULED)
PUT  /monthly-credits     -> HTTP 204   (was 404)
PUT  /billing-info        -> HTTP 204   (was 404)
DELETE (cancel)           -> HTTP 204   (was 404)
```

Backend `tsc --noEmit` clean. Note: 25 unrelated suites (RAG/artifacts/tools) fail to parse locally with a Jest ESM transform error; verified identical on the parent commit without these changes, so pre-existing and not introduced here.

Cancelling a scheduled subscription produces a `cancelledAt < startsAt` record — exactly the shape that used to 500 the history endpoint, which the parent PR's `getNextRenewalDate` fix handles. The two changes are complementary.

---

> [!NOTE]
> **Medium Risk**
> Touches super-admin billing mutations and cancellation event semantics; seat updates are explicitly unchanged to avoid licence bypass, but wrong event emission on cancel could still affect org-wide entitlements.
> 
> **Overview**
> Fixes super-admin **cancel**, **billing-info**, and **monthly-credits** mutations that returned `404` for **scheduled** subscriptions (`startsAt` in the future) because they resolved the org via `GetActiveSubscriptionUseCase` instead of the newest record the UI shows.
> 
> Those three flows now use **`findManageableSubscription`**, which picks the newest subscription by `createdAt` and still rejects orgs with more than one serving subscription. **`UpdateSeatsUseCase` stays on the active subscription** so org-admin invite auto-bump cannot raise seats on a future contract while the current licence stays full (documented + regression test).
> 
> **Cancel** only emits `SubscriptionCancelledEvent` when the subscription was actually serving, so cancelling a scheduled sub does not strip credit limits for a still-active one. **Uncancel** allows restoring subscriptions that were cancelled before their start date.
> 
> Adds unit coverage, module docs, E2E for scheduled subscription edit/cancel, and super-admin API client helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c759a177a293c109ce39fc876258edff99ffdaac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sirolad:
* **#1707** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1707?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1710, opened 2026-09-21
chore(deps): hold undici on the major node embeds (AYC-000)

sirolad:
## Why

Dependabot #1671 (`undici` 7.29.0 → 8.10.2) is not safe to merge **on the node image we run today**.

undici 8 removed the legacy dispatcher handler API (`onConnect`/`onHeaders`/`onData`/
`onComplete`) and now asserts the new one in `assertRequestHandler`. Node's built-in `fetch`
hands *its own embedded* undici's handler to whatever `dispatcher` you pass it, so the
dependency major and the runtime's embedded major have to agree:

| image | embeds undici |
| --- | --- |
| `node:24.16.0-alpine` (our pin) | 7.21.0 |
| `node:26` | 8.10.2 |

`McpOAuthFetchService` passes an `undici` `Agent` to `globalThis.fetch` to pin DNS lookups to
already-validated addresses (SSRF / DNS-rebinding guard). Mismatch the majors and it fails
before the connection is attempted:

```text
TypeError: fetch failed
  cause: Error: InvalidArgumentError: invalid onRequestStart method
    at makeNetworkError (node:internal/deps/undici/undici:10797:35)
```

Every MCP OAuth request would fail at runtime. CI caught it as
`mcp-oauth-fetch.service.spec.ts › blocks DNS rebinding between validation and connection`.

## What

- Ignore `undici` semver-major in Dependabot — same shape as the existing `node:26` ignore.
- **Cross-reference the two ignores**, because the trap is symmetric. The `node:26` ignore
  (AYC-423, 2026-07-13) predates this service (2026-08-04), so its "do not remove without
  re-validating" note says nothing about undici. Whoever lifts it and moves to `node:26`
  while undici stays on 7 gets the exact same breakage from the other direction.
- One comment at the import site, where a manual bump would start. Converted that file's one
  `../` import per the absolute-import rule.

## Exit condition

Not "someday" — Node 24 goes **maintenance-only on 2026-10-20** and Node 26 becomes Active
LTS on 2026-10-28. So the real fix is a coupled `node:26` + `undici@8` migration, which is
blocked on re-validating the AYC-423 connection stalls. This PR just makes sure the two pins
can't drift apart in the meantime. Follow-up ticket to be filed.

## Classification

Fast Path — Dependabot config + comments, no behavior change. Reversible, single layer.

## Validation

Worktree, Node 24.14.0, `ayunis-core-backend`:

| lockfile | `jest mcp-oauth-fetch.service.spec.ts` |
| --- | --- |
| #1671 (undici 8.10.2) | 1 failed, 9 passed — rebinding guard never reaches the network |
| this branch (undici 7.29.0) | 10 passed |

Root cause confirmed by walking the error `cause` chain, and against
`undici@8.10.2/lib/core/util.js:573`. Embedded versions from `process.versions.undici` (24)
and `nodejs/node@v26.x deps/undici/src/package.json` (26); LTS dates from
`nodejs/Release schedule.json`.

---

> [!NOTE]
> **Low Risk**
> Dependabot policy and documentation only; MCP OAuth fetch logic is unchanged aside from an import path refactor.
> 
> **Overview**
> Blocks Dependabot from proposing **semver-major `undici` upgrades** and documents that the `undici` major must stay aligned with Node’s embedded undici (Node 24 → undici 7; Node 26 → undici 8). The Docker `node` ignore comment now states that moving to `node:26` must include bumping `undici` to 8 in the same change.
> 
> In `McpOAuthFetchService`, adds an import-site note that the `undici` `Agent` used as `globalThis.fetch`’s `dispatcher` only works when majors match, and switches the port import to the project’s absolute `src/...` path. **No runtime behavior change**—this prevents a bad Dependabot merge that would break MCP OAuth fetches with `invalid onRequestStart method`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f8ba13ac26ec52f756779ee7453c85368d9e00f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sirolad:
* **#1710** <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ayunis-core/ayunis-core/1710?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

